PR #8, opened 2024-03-12
Feature/ic upload

JPHutchins:
This adds a routine for sending images via the upload command of the new Intercreate group.  It's tested at 127 MTU which seems to be the max supported by the Zephyr SMP Shell Transport.

### DIFF
--- a/smpclient/__init__.py
+++ b/smpclient/__init__.py
@@ -102,6 +102,9 @@ class SMPClient:
     def _cbor_integer_size(integer: int) -> int:
         """CBOR integers are packed as small as possible."""
 
+        # If the integer is less than 24, then the size is encoded in the same
+        # byte as the value.
+        # https://datatracker.ietf.org/doc/html/rfc8949#name-core-deterministic-encoding
         return 0 if integer < 24 else 1 if integer < 0xFF else 2 if integer < 0xFFFF else 4
 
     def _get_max_cbor_and_data_size(self, request: smpmsg.WriteRequest) -> Tuple[int, int]:
--- a/smpclient/__init__.py
+++ b/smpclient/__init__.py
@@ -1,10 +1,11 @@
 """Simple Management Protocol (SMP) Client."""
 
 from hashlib import sha256
-from typing import AsyncIterator, Final, cast
+from typing import AsyncIterator, Final, Tuple, cast
 
 from pydantic import ValidationError
 from smp import header as smpheader
+from smp import message as smpmsg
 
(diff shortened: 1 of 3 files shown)


smpclient/__init__.py
```diff
@@ -97,32 +98,53 @@ async def __aenter__(self) -> 'SMPClient':
     async def __aexit__(self) -> None:
         await self.disconnect()
 
-    def _maximize_packet(self, request: ImageUploadWrite, image: bytes) -> ImageUploadWrite:
-        """Given an `ImageUploadWrite` with empty `data`, return the largest packet possible."""
+    @staticmethod
+    def _cbor_integer_size(integer: int) -> int:
+        """CBOR integers are packed as small as possible."""
+
+        return 0 if integer < 24 else 1 if integer < 0xFF else 2 if integer < 0xFFFF else 4
```
swoisz, line 105: Maybe notify that 24 here is a significance for short field encoding? otherwise it looks like a magic number

smpclient/__init__.py
```diff
@@ -97,32 +98,53 @@ async def __aenter__(self) -> 'SMPClient':
...
+        return cbor_size, data_size
+
+    def _maximize_packet(self, request: ImageUploadWrite, image: bytes) -> ImageUploadWrite:
+        """Given an `ImageUploadWrite` with empty `data`, return the largest packet possible."""
 
-        chunk_size = self._transport.max_unencoded_size - len(bytes(request))
-        chunk_size -= cbor_integer_size(chunk_size)
-        chunk_size = max(0, chunk_size)
-        cbor_size = _h.length + chunk_size + cbor_integer_size(chunk_size)
+        h: Final = cast(smpheader.Header, request.header)
```
swoisz, line 133: Curious as to why `_h` was 'private' before?

swoisz:
image
Is this test coverage output expected?

smpclient/__init__.py
JPHutchins: I was using it as shorthand for "this is a temporary value" but decided it was more clear to just make it "h" for "header".

JPHutchins:
> image Is this test coverage output expected?

Yes 😢  - will be 100% eventually!

JPHutchins:
@swoisz  I just noticed that your terminal is not happy about the ANSI escapes in the terminal prefix - you should file a complaint!  https://github.com/JPHutchins/envr/issues

swoisz:
OK issue filed :D

swoisz (approved):
Looks great!

PR #10, opened 2024-03-15
feat: add shell management group

petrovichkr:
Add shell management group based on [Zephyr Docs](https://docs.zephyrproject.org/latest/services/device_mgmt/smp_groups/smp_group_9.html).


JPHutchins (approved):
Thank you!

PR #18, opened 2024-05-09
Fix instillation by pip

tomaszduda23:


pyproject.toml
```diff
@@ -1,3 +1,7 @@
+[project]
+name = "smpclient"
+version = "1.3.2"
```
JPHutchins, line 3: I'd prefer "0" if it is valid.  If not, then "0.0.0", or "0.1.0-dev", or anything to indicate that it is not a real version number.  This library uses the git tag as the SSOT for versioning so I'd rather give users that are installing unreleased versions no version rather than the wrong version.  Released version will always be available from PyPI.

JPHutchins (changes requested):

JPHutchins:
Hey @tomaszduda23 Can you try pip installing from this branch: https://github.com/intercreate/smpclient/tree/python/support-python3.8/9, e.g.
```
pip install git+https://github.com/intercreate/smpclient/@577a2a4c34ee014fef2a5c2ec90e422cff14bdc7a5c2ec90e422cff14bdc7
```
It seems to work for me.  Basically the pyyproject.toml was missing
```toml
[build-system]
requires = ["poetry-core>=1.0.0"]
build-backend = "poetry.core.masonry.api"
```

Thanks!

tomaszduda23:
> Hey @tomaszduda23 Can you try pip installing from this branch: https://github.com/intercreate/smpclient/tree/python/support-python3.8/9, e.g.
> 
> ```
> pip install git+https://github.com/intercreate/smpclient/@577a2a4c34ee014fef2a5c2ec90e422cff14bdc7a5c2ec90e422cff14bdc7
> ```
> 
> It seems to work for me. Basically the pyyproject.toml was missing
> 
> ```toml
> [build-system]
> requires = ["poetry-core>=1.0.0"]
> build-backend = "poetry.core.masonry.api"
> ```
> 
> Thanks!

It works :+1:

PR #25, opened 2024-05-26
feat: add serial transport line_length (default 128) and line_bufs (d…

### LINKED ISSUE #19: Image upload Write missing len parameter
I am testing the smpmgr and found a possible bug (?) when uploading an image to mcuboot over zephyr.

If the `ImageUploadWrite` packet does not include the `len` parameter, when the bootloader [decodes](
https://github.com/zephyrproject-rtos/mcuboot/blob/d4394c2f9b76e0a7b758441cea3a8ceb896f66c8/boot/boot_serial/src/boot_serial.c#L665) the `len` parameter it gets the maximum `uint32_t` number. This in turn makes mcuboot return `MGMT_ERR_EINVAL` and makes the upload from smpmgr fail as the offset field is [not sent](https://github.com/zephyrproject-rtos/mcuboot/blob/d4394c2f9b76e0a7b758441cea3a8ceb896f66c8/boot/boot_serial/src/boot_serial.c#L877) since rc == 3. 

Mcuboot fails because the first chunk uses the `len` field to check if the length of the image fits in the partition ([here](https://github.com/zephyrproject-rtos/mcuboot/blob/d4394c2f9b76e0a7b758441cea3a8ceb896f66c8/boot/boot_serial/src/boot_serial.c#L743)).

If I add the len field [here](https://github.com/intercreate/smpclient/blob/9279cd7a6c5e27579552df88f8ff4dbe78b0e5ec/smpclient/__init__.py#L81), then mcuboot complies and works correctly.  Something such as 
```python
            response = await self.request(
                self._maximize_packet(ImageUploadWrite(off=response.off, data=b''), image, len= len(image))
            )
```

I tested this with the latest mucboot repo from zephyr. In specific this [commit](https://github.com/zephyrproject-rtos/mcuboot/tree/d4394c2f9b76e0a7b758441cea3a8ceb896f66c8).

Smpmgr options

`upgrade zephyr.signed.hex`

### LINKED ISSUE #22: serial update do not work
I've checked 2.0.1 and serial update do not work. It seems that  following is missing.
```python

    async def send(self, data: bytes) -> None:
        logger.debug(f"Sending {len(data)} bytes")
        # line cannot be bigger than 128 bytes by default
        # https://github.com/zephyrproject-rtos/zephyr/blob/076c0df3d8a878e4457c1d831704cd33b9ead6d3/drivers/console/uart_mcumgr.c#L101
        # https://github.com/zephyrproject-rtos/zephyr/blob/076c0df3d8a878e4457c1d831704cd33b9ead6d3/drivers/console/Kconfig#L204
        for packet in smppacket.encode(data, line_length=128):
```

JPHutchins:
…efault 2); test Zephyr default USB SMP server example; refactor examples folder layout

This fixes some problems with USB/serial DFU defaults and compatibility.  Adds some examples of USB DFU to the main app.  Would like to add an example of a faster upload and an example of an MCUBoot single slot upload.

Fixes #22 
Fixes #19 (allow user to not send SHA)


smpclient/transport/serial.py
```diff
@@ -126,13 +161,9 @@ async def disconnect(self) -> None:
     @override
     async def send(self, data: bytes) -> None:
         logger.debug(f"Sending {len(data)} bytes")
```
JPHutchins, line 163: Should raise if len(data) > self.max_unencoded_size?

JPHutchins:
@tomaszduda23 I think this might be right.  LMK what you think.  The serial transport takes optional args for line length, line_bufs, and total size.  They get used in a few places to slice the upload chunks.

JPHutchins:
This is an NRF52840DK example: python -m examples.usb.upgrade nrf52840dk_nrf52840

I didn't make the same image for feather yet.

JPHutchins:
I've determined that these are the three settings that matter and correspond to line length, line buffers, and "max encoded size", respectively.  Examples are in the dutfirmware folder.
```
CONFIG_UART_MCUMGR_RX_BUF_SIZE=1024
CONFIG_UART_MCUMGR_RX_BUF_COUNT=1
CONFIG_MCUMGR_TRANSPORT_NETBUF_SIZE=1024
```

PR #35, opened 2024-07-15
File management

Maartenwn:
Added [file management](https://docs.zephyrproject.org/3.4.0/services/device_mgmt/smp_groups/smp_group_8.html#file-hash-checksum).

As this uses changes in [smp](https://github.com/JPHutchins/smp) that are currently not released this will be draft.

### DIFF
--- a/smpclient/__init__.py
+++ b/smpclient/__init__.py
@@ -14,6 +14,7 @@ from smp import message as smpmsg
 
 from smpclient.exceptions import SMPBadSequence, SMPUploadError
 from smpclient.generics import SMPRequest, TEr0, TEr1, TRep, error, success
+from smpclient.requests.file_management import FileUpload
 from smpclient.requests.image_management import ImageUploadWrite
 from smpclient.requests.os_management import MCUMgrParametersRead
 from smpclient.transport import SMPTransport
@@ -107,7 +108,7 @@ class SMPClient:
         """
 
         response = await self.request(
-            self._maximize_packet(
+            self._maximize_image_upload_write_packet(
                 ImageUploadWrite(
                     off=0,
                     data=b"",
@@ -133,7 +134,9 @@ class SMPClient:
         # send chunks until the SMP server reports that the offset is at the end of the image
         while response.off != len(image):
             response = await self.request(
-                self._maximize_packet(ImageUploadWrite(off=response.off, data=b""), image),
+                self._maximize_image_upload_write_packet(
+                    ImageUploadWrite(off=response.off, data=b""), image
+                ),
                 timeout_s=subsequent_timeout_s,
             )
             if error(response):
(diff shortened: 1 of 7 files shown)


JPHutchins:
Thanks for the detailed tests and examples!  Is there a standard Zephyr example that I can test with?

Seeing the download method as a regular call instead of a generator makes me realize that this should be an option for all uploads and downloads.

Maybe something like

download is a regular async function

download_iter is the async iterator (not sure I like that name)

download would wrap download_iter

And this approach would be for all download/upload methods, including firmware updates.

What do you think?

JPHutchins:
> Might be nice to check the checksum after uploading a file to check if the file is correctly uploaded.

That would be nice!  Like a default arg that will do the SMP calls asking for the checksum from the device, then compares with a checksum calculated from smpclient?

smpclient/__init__.py
```diff
@@ -154,6 +157,85 @@ async def upload(
...
+        first_timeout_s: float = 40.000,
+        subsequent_timeout_s: float = 2.500,
+    ) -> AsyncIterator[int]:
+        response = await self.request(
+            self._maximize_file_upload_packet(
+                FileUpload(name=file_path, off=0, data=b"", len=len(file_data)),
+                file_data,
+            ),
+            timeout_s=first_timeout_s,
+        )
```
JPHutchins, line 173: The `first_timeout_s` and `subsequent_timeout_s` parameters were informed by MCUmgr for FW uploads where the user has not configured "erase flash progressively" or "stream flash" and the device may take a while to do an erase on all the pages at once.

Let's start by assuming we don't need to do anything funny like that and just set one timeout: timeout_s = 2.500

Maartenwn:
> In the Zephyr tree, it looks like [samples/subsys/mgmt/mcumgr/smp_svr](https://github.com/zephyrproject-rtos/zephyr/tree/main/samples/subsys/mgmt/mcumgr/smp_svr) with `overlay-fs.conf` would be the closest match.

Looks the best however that example doesn't enable hashing by default. This can be enabled by enabling the MCUMGR_GRP_FS_CHECKSUM_HASH. You also have to enable an hashing algoritm by setting MCUMGR_GRP_FS_CHECKSUM_IEEE_CRC32 and/or MCUMGR_GRP_FS_HASH_SHA256 (this requires also setting an backend),  see https://docs.zephyrproject.org/latest/services/device_mgmt/smp_groups/smp_group_8.html#file-hash-checksum

JPHutchins:
Hello!  The breaking changes from SMP 3.0.1 have been merged to main as of e5727fb8caa5a9c4bdb3e1487bd7f8258526b82d.  This branch can be rebased to integrate the changes.  Primarily it's renaming ErrorV0 -> ErrorV1 and ErrorV1 -> ErrorV2.

Maartenwn:
> > Might be nice to check the checksum after uploading a file to check if the file is correctly uploaded.
> 
> That would be nice! Like a default arg that will do the SMP calls asking for the checksum from the device, then compares with a checksum calculated from smpclient?

I am not sure if this belongs in this pr, as this could be in a separate pr.

JPHutchins (approved):
Thank you!  The test cases are really good!

JPHutchins:
> I am not sure if this belongs in this pr, as this could be in a separate pr.

Agreed, we'll move along.

PR #58, opened 2024-12-10
breaking: remove address from SMPClient; SMPTransport owns address

JPHutchins:
I've long thought that passing the SMP server "address" to the SMPClient class instead of the SMPTransport implementation classes was a mistake. "address" means different things to the different transports. Notably, this change allows the UDP transport to clearly take an address and a port argument, rather than just an address.

Unfortunately this is a breaking change. While all existing address args should still be valid, they must be moved from "1st arg of SMPClient() to 1st arg of SMPTransport".

@M1cha this is what I have in mind as the breaking change required to continue implementing the unique requirements of each transport for smpmgr.

### DIFF
--- a/smpclient/transport/ble.py
+++ b/smpclient/transport/ble.py
@@ -68,7 +68,14 @@ logger = logging.getLogger(__name__)
 class SMPBLETransport(SMPTransport):
     """A Bluetooth Low Energy (BLE) SMPTransport."""
 
-    def __init__(self) -> None:
+    def __init__(self, name_or_address: str) -> None:
+        """Initialize the SMP BLE transport.
+
+        Args:
+            name_or_address: The device name or address (MAC or UUID) to connect to.
+        """
+        self._name_or_address: Final = name_or_address
+
         self._buffer = bytearray()
         self._notify_condition = asyncio.Condition()
         self._disconnected_event = asyncio.Event()
@@ -80,12 +87,13 @@ class SMPBLETransport(SMPTransport):
         logger.debug(f"Initialized {self.__class__.__name__}")
 
     @override
-    async def connect(self, address: str, timeout_s: float) -> None:
-        logger.debug(f"Scanning for {address=}")
+    async def connect(self, timeout_s: float) -> None:
+        logger.debug(f"Scanning for {self._name_or_address=}")
         device: BLEDevice | None = (
-            await BleakScanner.find_device_by_address(address, timeout=timeout_s)
-            if MAC_ADDRESS_PATTERN.match(address) or UUID_PATTERN.match(address)
-            else await BleakScanner.find_device_by_name(address)
+            await BleakScanner.find_device_by_address(self._name_or_address, timeout=timeout_s)
+            if MAC_ADDRESS_PATTERN.match(self._name_or_address)
+            or UUID_PATTERN.match(self._name_or_address)
+            else await BleakScanner.find_device_by_name(self._name_or_address)
         )
 
         if type(device) is BLEDevice:
@@ -95,7 +103,7 @@ class SMPBLETransport(SMPTransport):
                 disconnected_callback=self._set_disconnected_event,
             )
         else:
-            raise SMPBLETransportDeviceNotFound(f"Device '{address}' not found")
+            raise SMPBLETransportDeviceNotFound(f"Device '{self._name_or_address}' not found")
 
         logger.debug(f"Found device: {device=}, connecting...")
         await self._client.connect()
--- a/examples/ble/helloworld.py
+++ b/examples/ble/helloworld.py
@@ -16,7 +16,7 @@ async def main() -> None:
     print(f"Found {len(smp_servers)} SMP servers: {smp_servers}")
 
     print("Connecting to the first SMP server...", end="", flush=True)
-    async with SMPClient(SMPBLETransport(), smp_servers[0].address) as client:
+    async with SMPClient(SMPBLETransport(smp_servers[0].address)) as client:
         print("OK")
 
         print("Sending request...", end="", flush=True)
--- a/examples/ble/imagestate.py
+++ b/examples/ble/imagestate.py
@@ -16,7 +16,7 @@ async def main() -> None:
     print(f"Found {len(smp_servers)} SMP servers: {smp_servers}")
 
     print("Connecting to the first SMP server...", end="", flush=True)
-    async with SMPClient(SMPBLETransport(), smp_servers[0].address) as client:
+    async with SMPClient(SMPBLETransport(smp_servers[0].address)) as client:
         print("OK")
 
         print("Sending request...", end="", flush=True)
--- a/examples/ble/mcumgrparameters.py
+++ b/examples/ble/mcumgrparameters.py
@@ -16,7 +16,7 @@ async def main() -> None:
     print(f"Found {len(smp_servers)} SMP servers: {smp_servers}")
 
     print("Connecting to the first SMP server...", end="", flush=True)
-    async with SMPClient(SMPBLETransport(), smp_servers[0].address) as client:
+    async with SMPClient(SMPBLETransport(smp_servers[0].address)) as client:
         print("OK")
         print(f"Client MTU is {client._transport.mtu}B")
         print(f"Client max unencoded size is {client._transport.max_unencoded_size}B")
--- a/examples/ble/upgrade.py
+++ b/examples/ble/upgrade.py
@@ -64,7 +64,7 @@ async def main() -> None:
     print("OK")
 
     print("Connecting to A SMP DUT...", end="", flush=True)
-    async with SMPClient(SMPBLETransport(), a_smp_dut.name or a_smp_dut.address) as client:
+    async with SMPClient(SMPBLETransport(a_smp_dut.name or a_smp_dut.address)) as client:
         print("OK")
 
         async def ensure_request(request: SMPRequest[TRep, TEr1, TEr2]) -> TRep:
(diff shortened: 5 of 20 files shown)


smpclient/transport/ble.py
```diff
@@ -80,12 +87,13 @@ def __init__(self) -> None:
         logger.debug(f"Initialized {self.__class__.__name__}")
 
     @override
-    async def connect(self, address: str, timeout_s: float) -> None:
-        logger.debug(f"Scanning for {address=}")
+    async def connect(self, timeout_s: float) -> None:
```
M1cha, line 90: This PR didn't change the behavior, but it got me wondering: What is the reasoning behind having a single parameter for both name and address? While this is probably fine for 99% of use-cases, you can easily make this connect to the wrong device by giving a device a name, that looks the same as the BLE address of another device :see_no_evil: :sweat_smile:

M1cha (approved):
I didn't test this, yet but it looks good, thx.

smpclient/transport/ble.py
JPHutchins: Yeah, I don't love it, and thought about creating a union type that's name or address, but honestly, if someone uses a MAC address as the device name, I WANT them to open a ticket - force them out into the wild! 🤣

GHF:
> I've long thought that passing the SMP server "address" to the SMPClient class instead of the SMPTransport implementation classes was a mistake.

I really like the direction this is going and I want to suggest that this could go even further. In this case, it's SMPClient dictating to SMPTransport what its connection parameters should be, when the transport-specific class would know better. I think the lifetime of a transport's connection session should be decoupled from SMPClient's session too.

We've been integrating smpclient into an application that has a standing BLE link and the semantics of SMPClient (the class and async context) owning the _lifetime_ of the transport's session has been awkward. It wants to make a BLE connection, but the establishment procedure isn't idempotent (scanning fails if the peripheral stops advertising with an existing link), and it disconnects the link through bleak when the SMPClient context closes.

Instead, if the semantics were: an SMPTransport instance exists only after a session is established and SMPClient can assume that a transport is available to send and receive as long as it's open, then it becomes more flexible to reuse a BLE link.

Ideally it'd probably look something like:

```py
async with SMPBLETransport(address) as transport:
	async with SMPClient(transport) as client:
        ...
```

JPHutchins:
> > I've long thought that passing the SMP server "address" to the SMPClient class instead of the SMPTransport implementation classes was a mistake.
> 
> I really like the direction this is going and I want to suggest that this could go even further. In this case, it's SMPClient dictating to SMPTransport what its connection parameters should be, when the transport-specific class would know better. I think the lifetime of a transport's connection session should be decoupled from SMPClient's session too.
> 
> We've been integrating smpclient into an application that has a standing BLE link and the semantics of SMPClient (the class and async context) owning the _lifetime_ of the transport's session has been awkward. It wants to make a BLE connection, but the establishment procedure isn't idempotent (scanning fails if the peripheral stops advertising with an existing link), and it disconnects the link through bleak when the SMPClient context closes.
> 
> Instead, if the semantics were: an SMPTransport instance exists only after a session is established and SMPClient can assume that a transport is available to send and receive as long as it's open, then it becomes more flexible to reuse a BLE link.
> 
> Ideally it'd probably look something like:
> 
> ```python
> async with SMPBLETransport(address) as transport:
> 	async with SMPClient(transport) as client:
>         ...
> ```

I think that I'm understanding that with this design the connected transport is an invariant of the SMPClient instance. This is nice, because it does away with SMPClient having a connect method and skirts around the general issue of python not having an `__ainit__()` constructor. It allows for client errors to occur without ending the transport lifetime, but semantically shows that a transport lifetime must live at least as long as a client.

Am I understanding correctly?

JPHutchins:
> [!WARNING]
> **LLM Disclosure**
>
> This comment was authored by claude-opus-4-8 on behalf of @JPHutchins, coordinating this breaking change with the cross-repo migration.

Staging note for the transport-subcommands epic (intercreate/smpmgr#103): breaking work lands on a per-repo `screaming-goblin` branch, and `smpclient`'s `screaming-goblin` will **pin `smp`'s `screaming-goblin`** (git dependency) so we can migrate smp → smpclient → smpmgr in lockstep before `smp` cuts its breaking release; the pin then moves to the released version.

As the main breaking prerequisite, this PR belongs on that `screaming-goblin` branch rather than `main` — i.e. retarget the base to `smpclient`'s `screaming-goblin` when we cut it (the `BREAKING/move-address-to-transport` branch itself is unchanged). That keeps `main` releasable while the breaking set assembles alongside the `smp` Frame[T]/Data + generics migration (JPHutchins/smp#65, JPHutchins/smp#62).

JPHutchins:
> [!WARNING]
> **LLM Disclosure**
>
> This comment was authored by `claude-opus-5-5[1m]` on behalf of @JPHutchins. We iterated on the design for this change over several rounds; @JPHutchins asked me to post the result here before implementing, and to close this PR as superseded.

**The design has grown past this PR, so I'm closing it as superseded.** The implementation will be one new PR on `screaming-goblin`, per the epic (intercreate/smpmgr#103). @GHF @M1cha: this is the shape we're building, and feedback is welcome during review.

```python
async with (
    SMPSerialTransport.connected("/dev/ttyACM0") as transport,  # opens the link; closes it on exit
    SMPClient(transport) as client,                            # negotiates sizing if the transport asks
):
    await client.upload(image)

# @GHF's standing link: smpclient never establishes or drops it
async with SMPBLETransport.borrow(bleak_client) as transport:
    async with SMPClient(transport) as client:
        ...
```

- **The transport owns its address, its options, and its link.** Each transport's constructor wraps an *already-open* resource, so an unconnected transport can't be held. `.connected(...)` opens one; `.borrow(...)` wraps one you already own, subscribing on entry and unsubscribing on exit.
- **`SMPClient` holds a type that cannot close the link.** `connect`/`disconnect` leave the `SMPTransport` Protocol, and `SMPClient.connect()`, `.disconnect()` and `.address` go away.
- **MCUmgr params negotiation is the transport's policy, and it's opt-in.** Each transport declares a sizing sum type. The client's bracket reads the params only when the transport was built with `Auto()`.

<details>
<summary>@GHF's standing-link requirements, and how this meets them</summary>

From https://github.com/intercreate/smpclient/pull/58#issuecomment-2848208698:

1. **A caller-owned BLE link backs an `SMPClient` without smpclient establishing it.** `SMPBLETransport.borrow(bleak_client)`, and the same for `SMPBumbleTransport.borrow(connection)`.
2. **The client going out of scope never drops the link.** `SMPClient` holds an `SMPTransport`, and nothing on that Protocol opens or closes a link. Reaching for one is a type error under mypy and pyright.
3. **Two clients can use one link in sequence without re-establishing it.** Negotiation is idempotent transport state.
4. **The link lives at least as long as the client.** The nesting of the brackets enforces it.

</details>

<details>
<summary>Sizing: per-transport sum types sharing one <code>Auto()</code></summary>

The server's MCUmgr `buf_size` is one global buffer size, but each transport turns it into a message cap differently. Zephyr `main` @ `70be2ff0`: [os_mgmt.c](https://github.com/zephyrproject-rtos/zephyr/blob/70be2ff0b565a3313128f5577f51cfeb3ebcf602/subsys/mgmt/mcumgr/grp/os_mgmt/src/os_mgmt.c#L551-L554).

| Transport | Sizing | `Auto()` gives |
|---|---|---|
| console serial | `Auto \| BufferSize \| BufferParams` (unchanged) | `buf_size - 4` (the length and CRC16 share the buffer) |
| raw serial, COBS | `Auto \| BufferSize` | `buf_size` |
| BLE (bleak, bumble) | `Auto \| Unfragmented \| BufferSize` | `buf_size`, fragmented across writes. This assumes the server enables `CONFIG_MCUMGR_TRANSPORT_BT_REASSEMBLY`, as Zephyr's `smp_svr` sample does. `Unfragmented()` means one message per write, for servers that don't reassemble. |
| UDP | `Auto \| BufferSize` | `min(buf_size, MSS)`, since the server receives each request as one datagram into one buffer (the same fix is going to `main` in #142) |

The mechanism: `SMPClient.__aenter__` calls `await transport.negotiate(read)`, where `read` sends the params request. Only the transport's `Auto()` arm calls `read()`, inside an exhaustive `match` over its sizing type. So the policy is decided in one place, and a pinned size never costs a round trip. If the server answers with an error (no params support), the transport warns and falls back to its default size. If it doesn't answer at all, `TimeoutError` propagates.

</details>

<details>
<summary>Transport options: #90 and #103</summary>

Options move onto each transport's `.connected(...)`. For BLE they are bleak's own backend argument types, passed through unchanged:

- #90 (cached GATT services on Windows, from intercreate/smpmgr#82): `winrt=WinRTClientArgs(use_cached_services=True)`.
- #103 (choose the HCI adapter): `bluez=BlueZClientArgs(adapter="hci1")`, applied to both the scan and the connection.
- UDP takes `address` *and* `port`, instead of `port` arriving through a `connect()` kwarg the Protocol never declared. That was this PR's original motivation.

</details>

<details>
<summary>Breaking changes</summary>

- `SMPClient(transport, address, ...)` becomes `SMPClient(transport, ...)`, and is an `async with` bracket that does no link I/O. `SMPClient.connect()`, `.disconnect()` and `.address` are removed.
- Transports are obtained from `Transport.connected(...)` or `.borrow(...)` rather than `Transport(...)` followed by `connect(address, timeout_s)`. The connect timeout moves to `.connected(...)`, and `SMPClient`'s `timeout_s` remains the request timeout.
- `SMPTransport` loses `connect`/`disconnect`, and `initialize(buf_size)` becomes `negotiate(read)`.
- For a transport sized with `Auto()`, a params timeout now raises; it used to be logged and ignored.
- No deprecation aliases: `screaming-goblin` is the breaking release.

</details>

JPHutchins:
> ```python
> async with (
>     SMPSerialTransport.connected("/dev/ttyACM0") as transport,  # opens the link; closes it on exit
>     SMPClient(transport) as client,                            # negotiates sizing if the transport asks
> ):
>     await client.upload(image)
> 
> # @GHF's standing link: smpclient never establishes or drops it
> async with SMPBLETransport.borrow(bleak_client) as transport:
>     async with SMPClient(transport) as client:
>         ...
> ```

My only note here is that the double context manager is only required if negotiating mcumgr params. Yet we're forcing everything to use it.

PR #59, opened 2024-12-10
fix: udp: use mtu as max_unencoded_size

### LINKED ISSUE #57: MTU size handling
Currently, the code tries to use "MCUMgrParametersRead" and the returned max buffer size as the max MTU, if supported. 

But look at one of zephyrs MTU kconfigs:
```
config MCUMGR_TRANSPORT_UDP_MTU
	int "UDP SMP MTU"
	default 1500
	help
	  Maximum size of SMP frames sent and received over UDP, in bytes.
	  This value must satisfy the following relation:
	  MCUMGR_TRANSPORT_UDP_MTU <= MCUMGR_TRANSPORT_NETBUF_SIZE + SMP msg overhead - address size
	  where address size is determined by IPv4/IPv6 selection.
```

The other transports have similar formulas, where MCUMGR_TRANSPORT_NETBUF_SIZE is always bigger than the transport MTU. In other words: They treat the buffer size as a maximum internal buffer, which has to be >= the largest MTU size of any transport.

This currently makes smpclient unable to use UDP transports. The size returned by the parameter API might be something large like 2048, but that doesn't mean that such large UDP packets are supported(which currently defaults to 1500). You might also want to use smaller MTUs either way to reduce the need for fragmentation, e.g. when doing UDP over a thread network.

My current workaround is to disable support for the mcumgr parameter API, so it has to fall back to the user-provided value, which can be set via smpmgrs CLI for example.

A possible solution might be to make the code prefer the transport-provided MTU over the reported buffer size. What do you think?

JPHutchins:
Fixes #57

### DIFF
--- a/smpclient/transport/udp.py
+++ b/smpclient/transport/udp.py
@@ -99,4 +99,9 @@ class SMPUDPTransport(SMPTransport):
     @override
     @property
     def mtu(self) -> int:
-        return self._smp_server_transport_buffer_size or self._mtu
+        return self._mtu
+
+    @override
+    @property
+    def max_unencoded_size(self) -> int:
+        return self._mtu


M1cha (approved):
I tested it and it works, thanks.

PR #61, opened 2025-01-30
Bugfix incorrect header lenght, with packet of data size 0xFF or 0xFFFF 

Maartenwn:
Fixed a bug causing the header length to be incorrect when the last packet had a size of 0xFF or 0xFFFF. The cbor_integer_size function returned 1 bytes to much for the value of 0xFF and 0xFFFF.


JPHutchins (approved):
Amazing find! I'm glad that this sketchy "cbor integer size" function is getting more coverage.

PR #64, opened 2025-05-02
fix: SMPClient context logs formatted traceback, not repr

GHF:
Previously, exiting the SMPClient async context with an exception results in logging f"{traceback=}" whose output looks like:

> `traceback=<traceback object at 0x7f4c3fea2980>`

as part of a full log message like so:

> `Exception in SMPClient: exc_type=<class 'smpclient.transport.SMPTransportDisconnected'>, exc_value=SMPTransportDisconnected('SmpNativeTransport cancelled receive due to disconnection'), traceback=<traceback object at 0x7f4c3fea2980>`

Instead, format the exception using traceback.format_exc() (which also grabs the originating exception with sys.exception()). Now the output looks like:

> ```
>   Exception in SMPClient:
>   Traceback (most recent call last):
>     File "<string>", line 11, in run_echo_test
>     File "/opt/axe/develop/latest/desktop/sysroots/x86-64-v3-poky-linux/usr/lib/python3.12/site-packages/smpclient/__init__.py", line 160, in request
>       frame = await self._transport.send_and_receive(request.BYTES)
>               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>     File "<string>", line 67, in send_and_receive
>     File "<string>", line 62, in receive
>   smpclient.transport.SMPTransportDisconnected: SmpNativeTransport cancelled receive due to disconnection
> ```

### DIFF
--- a/smpclient/__init__.py
+++ b/smpclient/__init__.py
@@ -27,6 +27,7 @@ from __future__ import annotations
 
 import asyncio
 import logging
+import traceback
 from hashlib import sha256
 from types import TracebackType
 from typing import AsyncIterator, Final, Tuple, Type
@@ -401,10 +402,10 @@ class SMPClient:
         self,
         exc_type: Type[BaseException] | None,
         exc_value: BaseException | None,
-        traceback: TracebackType | None,
+        tb: TracebackType | None,
     ) -> None:
         if exc_value is not None:
-            logger.error(f"Exception in SMPClient: {exc_type=}, {exc_value=}, {traceback=}")
+            logger.error(f"Exception in SMPClient:\n{traceback.format_exc()}")
         await self.disconnect()
 
     @staticmethod


smpclient/__init__.py
```diff
@@ -401,10 +402,10 @@ async def __aexit__(
         self,
         exc_type: Type[BaseException] | None,
         exc_value: BaseException | None,
-        traceback: TracebackType | None,
+        tb: TracebackType | None,
     ) -> None:
         if exc_value is not None:
-            logger.error(f"Exception in SMPClient: {exc_type=}, {exc_value=}, {traceback=}")
+            logger.error(f"Exception in SMPClient:\n{traceback.format_exc()}")
         await self.disconnect()
```
JPHutchins, line 409: What do you think about `raise exc_value` after the disconnect? I wonder what the precedent is for async context managers?

JPHutchins (approved):
Thanks! This is a great improvement. Getting good exceptions from asyncio is always challenging, and I'm not certain of best practices. What do you think about raising the exception after the disconnect so that users can catch them?

I read through this conversation recently, but I'm not sure that I absorbed it!

https://discuss.python.org/t/asyncio-tasks-and-exception-handling-recommended-idioms/23806

smpclient/__init__.py
JPHutchins: I suppose that if it raises, then we should not log the traceback and leave that up to the caller.

JPHutchins: This would interact with the other discussion, since the disconnect() method would go away.

GHF: > What do you think about `raise exc_value` after the disconnect? I wonder what the precedent is for async context managers?

That was actually my initial desired change, but it's also a breaking API change and I don't know how much other users of the library would tolerate that, so I'll merge this as an easy incremental improvement.

Besides, I didn't know how much change in this area I wanted was because of the war crimes I'm doing or if it was generally applicable: I'm embedding an interpreter into a C++ application where my `SMPTransport` is a C++/Python bindings-bound `SmpNativeTransport` class (hence the other discussion) and logging is redirecting through a custom `logging.Handler` added to the appropriate `logging.getLogger(<names I care about>)`.

The logging situation, plus the "eating" of exceptions in this coro, meant that to get tracebacks for my `SmpNativeTransport` I needed to fix this .error log site and add my `Handler` to `logging.getLogger('smpclient')`, which is not a big deal but took me a fair bit of empty-output crashes to figure out.

So you can see I'm a bit reluctant to make suggestions about the direction for the SMPClient API in general because I'm using it in such an unusual way.

JPHutchins: I need to test. Since I sorta expect an exception that's raised in the context manager to be raised on exit. If python doesn't do that, I want to understand why, then decide to log error instead of raising. My preference is for raising. IMO it's not a breaking change since user's apps are broken without them knowing!

Initial reading of this is that it should be raising on exit unless `__aexit__` returns `True`.

Source: https://bbc.github.io/cloudfit-public-docs/asyncio/asyncio-part-3.html

JPHutchins (approved):
I've looked into this and it seems like suppressing exceptions is the expected behavior for async context managers. The exception cannot be (easily) retrieved outside of the block because the context has ended.

So, logging will have to do for now!

PR #65, opened 2025-05-23
fix: Handle Mac Sequoia UUIDs

ball-hayden:
Consider the following UUID, produced by Mac Sequoia:

```
39806566-8AB0-E619-5A75-79E65ADE7105
```

This should be handled as a UUID, rather than a device name.


tests/test_smp_ble_transport.py
```diff
@@ -65,11 +65,6 @@ def test_UUID_PATTERN() -> None:
     assert UUID_PATTERN.match(UUID("0000FFFF-0000-4FFF-a000-FFFFFFFFFFFF").hex)
     assert UUID_PATTERN.match(UUID("FFFF0000-FFFF-4000-bFFF-000000000000").hex)
 
-    assert not UUID_PATTERN.match("00000000-0000-0000-8000-000000000000")
-    assert not UUID_PATTERN.match("FFFFFFFF-FFFF-1FFF-9FFF-FFFFFFFFFFFF")
-    assert not UUID_PATTERN.match("0000FFFF-0000-4FFF-c000-FFFFFFFFFFFF")
-    assert not UUID_PATTERN.match("FFFF0000-FFFF-4000-dFFF-000000000000")
```
ball-hayden, line 71: While, strictly, these are not valid UUIDs, they are UUID-like enough that they should probably be treated that way?

Otherwise, perhaps we could lean on `uuid`'s functionality to assert whether a UUID is valid rather than using regex?

ball-hayden:
@JPHutchins I think this is ready for your review now, please - sorry for the earlier test failures.

tests/test_smp_ble_transport.py
JPHutchins: It looks like I've used a too narrow definition for UUID. The specific spec that it's enforcing is here - see the 4 and 8 in the version and variant fields: https://www.rfc-editor.org/rfc/rfc9562#name-version-field

I'd love to get rid of the regex in this PR if you think that's for the best - are you willing to give it a shot and update the tests?

Concerns have previously been raised about using strings for the 3 "address" kinds:

- name
- MAC
- UUID

A more precise definition of the argument could be as a sum type.

```python
name: AdvertisedName # type alias str
mac: MACAddress # define a type that acts like UUID built in or find a built in
uuid: UUID # built in
```

And then app layer, eg smpmgr, decides how it wants to deal with raw strings.

JPHutchins:
> @JPHutchins I think this is ready for your review now, please - sorry for the earlier test failures.

Thanks for fixing this!

tests/test_smp_ble_transport.py
ball-hayden: e.g:

```python
@override
    async def connect(
        self,
        timeout_s: float,
        address: str | None = None,
        uuid: str | None = None,
        name: str | None = None,
    ) -> None:
```

or, even, handing off discovery of the device to the user entirely:

```python
async def connect(self, timeout_s: float, device: BLEDevice) -> None:
```

JPHutchins: Great points! Let's keep it simple here and avoid the breaking changes.

There's a list of upcoming breaking changes, particularly moving the "address" argument to the Transport instead of the Client. That creates a good opportunity for improving transport specific parameters.

JPHutchins (approved):
Thank you!

Issue #63, opened 2025-04-01
Update to poetry 2.0.0


otavio:
Is it possible for you to do the same upgrade here and update the build system to the newer version of it? I'm asking because I need that to include inside of the Nix ecosystem.

JPHutchins:
Yes, I'd like to do that. I'm interested in using Nix, particularly to replace Yocto, so I am curious to learn more.

Why is the Nix build system not getting the compiled wheel that smpmgr specifies? That is, why is it bothering to compile this package at all?

https://pypi.org/project/smpclient/#files

otavio:
What happens is that it is fully reproducible, so it uses the predefined recipes to build the package and not fetch them directly from the Python servers. So the version that it has available is the 2.0.0 and consequently it has the newer plugins as well.

Take a look at the documentation below about the introduction and how it works and what is the benefits it brings:

https://nixos-and-flakes.thiscute.world/introduction/

and how it can be used to leverage the ecosystem to provide the two-chain environment for the Zypher use:

https://github.com/OSSystems/ZephyrBT

Regarding the Yocto project, depending on the use case that you have, it may or may not be an option. Specifically for bigger systems, it is appealing because it provides the customer's ability without the need of you working in the recipes yourself. On the other hand, for very small systems, I think that is challenging because the store tends to not be so small and eventually you may have multiple versions of the same packages included in your system.

We are looking at using the Nix for our build system in some robotics devices which we are going to use it to integrate the ROS2 framework.

JPHutchins:
@otavio Sorry for delay here!

As an aside, sounds like Nix could stand to improve their PEP517 compat. E.g., if a package is PEP 517 compliant, nix should be able to build it.

otavio:
It has built it. However, it didn't get the version correctly because you are using the version of the plugin that is not compatible with the newer Poetry version. I have already packaged it, but it would be great if you could update that anyway.

PR #68, opened 2025-05-31
BREAKING: drop Python 3.8 support; update poetry and lock

JPHutchins:
Bleak dropped python 3.8 support, so we're following. It is only a breaking change for downstream projects using python 3.8. Otherwise, updating to poetry 2 and poetry-dynamic-versioning, plus fixing the PyPI flow.


JPHutchins:
Closes #63

PR #71, opened 2025-07-28
feat: add enumeration management group

sgfeniex:
Adds the enumeration management group to this layer

related: https://github.com/JPHutchins/smp/pull/40

https://github.com/JPHutchins/smp/pull/44 is a required fix that enables parsing `GroupDetailsResponse`

### DIFF
--- a/smpclient/requests/enumeration_management.py
+++ b/smpclient/requests/enumeration_management.py
@@ -1,0 +1,22 @@
+from smp import enumeration_management as smpem
+
+
+class _EnumGroupBase:
+    _ErrorV1 = smpem.EnumManagementErrorV1
+    _ErrorV2 = smpem.EnumManagementErrorV2
+
+
+class CountSupportedGroups(smpem.GroupCountRequest, _EnumGroupBase):
+    _Response = smpem.GroupCountResponse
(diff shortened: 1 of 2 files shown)


JPHutchins:
> > Looks like we might need a change in `smp` as well. `GroupDetails.id` might need to become `GroupDetails.group`. Checking Zephyr source and docs on this.
> > ...seems to check out.
> > https://docs.zephyrproject.org/latest/services/device_mgmt/smp_groups/smp_group_10.html#details-on-supported-groups-request
> 
> I agree, and that's a welcome change to the API:
> 
> https://github.com/JPHutchins/smp/blob/085c8b17e0af7f76f5b861c202f54c6306efab0c/smp/enumeration_management.py#L107-L117

Fix is in SMP 3.3.2: https://pypi.org/project/smp/

Note: I think that the 3 repo layout is not worth it for dynamic versioning. Then again, 3 python packages in one repo... I'd need to see an example 🤣.

JPHutchins:
@sgfeniex FYI that there will be a (good) breaking change upstream in SMP soon: https://github.com/JPHutchins/smp/issues/45

PR #82, opened 2025-10-23
fix: ble: Race condition in receive for notifications

randyscott:
Fix a race condition relating to BLE notifications. If the notification arrives before the `receive()` method is called, the notification will be lost and the transfer will time out. This happens because `self._notify_condition.notify()` does nothing without a pending `self._notify_condition.wait()`.

Only wait for notifications if the buffer does not contain enough data for the header.

### DIFF
--- a/smpclient/transport/ble.py
+++ b/smpclient/transport/ble.py
@@ -159,13 +159,9 @@ class SMPBLETransport(SMPTransport):
         #       self._notify_condition is used to synchronize access to self._buffer.
 
         async with self._notify_condition:  # wait for the header
-            logger.debug(f"Waiting for notify on {SMP_CHARACTERISTIC_UUID=}")
-            await self._notify_or_disconnect()
-
-            if len(self._buffer) < smphdr.Header.SIZE:  # pragma: no cover
-                raise SMPBLETransportException(
-                    f"Buffer contents not big enough for SMP header: {self._buffer=}"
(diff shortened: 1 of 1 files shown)


JPHutchins:
> > I think that this makes sense - does it happen when using the request method which normally calls send_and_receive?
> 
> 
> 
> Yes, exactly. This was an image upload, uses `send_and_receive`. Was able to catch a debug log showing the notification arriving before the wait. This was the cause of pretty consistent failures when programming one of my boards.
> 
> 
> 
> > If we had to, we could start waiting for the response before sending the request. Or have a receive task always running, carefully.
> 
> 
> 
> Would be a bigger change, of course. I do like the simplicity of this change here and it seems like a pretty safe way to handle the race condition.
> 
> 

I was worried that the fix is relying on subsequent bytes being received, but looking again I think that it's not a problem. If I understand correctly, in the original error case of response arriving before wait(), this fix means that the wait call never happens (good). Is that correct? Is the same bug present in the code that receives the rest of the packet?

I found this feature, and I'm curious if it would help: https://docs.python.org/3/library/asyncio-sync.html#asyncio.Condition.wait_for

If not, I'm happy to merge as is.

Thanks again for finding and fixing this! I'll bump smpclient and smpmgr as soon as this is merged.

JPHutchins (approved):

Issue #12, opened 2024-03-29
Improve the implementation that handles non-SMP traffic on the serial transport


JPHutchins:
This PR added "support" for Zephyr sending log messages in between SMP packets: https://github.com/intercreate/smpclient/commit/31717db7ef7e65e30c65142dcfb87dc24e3456dd

The implementation and testing was rushed and of low quality, in my opinion.

Further, I do not like that messages from Zephyr are logged.  I feel that they should be returned to the caller and the caller decides what to do with them.  The problem is that it is only the SMP Shell Transport that has this issue.  One way to handle it would simply be to fill an asyncio queue with the logs (non SMP messages) and then the caller can simply choose whether or not to clear the queue.

PR #87, opened 2025-11-29
Cleanup serial data handling

JaagupAverin:
chore: Cleanup non-smp serial data handling.

Makes the buffer processing hopefully a bit more clear. Non-SMP data is no longer logged to stdout but can be accessed with the read_serial() function.

### DIFF
--- a/smpclient/transport/serial.py
+++ b/smpclient/transport/serial.py
@@ -42,30 +42,17 @@ class SMPSerialTransport(SMPTransport):
     _POLLING_INTERVAL_S = 0.005
     _CONNECTION_RETRY_INTERVAL_S = 0.500
 
-    class _ReadBuffer:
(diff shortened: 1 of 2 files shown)


JPHutchins:
Thanks! This should resolve #12, take a look for details. I hope it's possible to avoid creating tasks and rely on queues instead. Callback could be useful for flexibility, but I suspect that the average use case would be to await the device log "channel" - the queue.

The refactor does appear to simplify the code, nice work!

JaagupAverin:
I think a lot of use cases are going to need a separate task for handling the serial data anyways so I wanted to hide it underneath the library. But I understand you'd rather not have this complexity. Made quick force push with new implementation.
Will be testing and trying this out a bit more thoroughly in the next week myself.

PR #91, opened 2025-12-18
Jp/fix linter (ignore)

JPHutchins:


JPHutchins:
> Looks good, thanks.
> Personally I like 'raise from none' [pattern](https://www.bugsink.com/blog/using-raise-from-none-in-python/) to reduce error clutter, but it's up to you.

That is nice, I do think it's a good fit here, feel free to change it back in the PR. Now, if only we could just remove exceptions from the language and call it a day!

JPHutchins:
Closed in favor of #116

PR #94, opened 2025-12-19
refactor: use a single configurable timeout value

etiennedm:
This is adressing https://github.com/intercreate/smpmgr/issues/71. Replace magic default values to use a single per-client default timeout value. This allows higher layers (e.g. smpmgr) to specify a timeout value when necessary and otherwise rely on the default set when creating the client.

I removed a bunch of default values that were all over the place, please comment on the need to have these, I might be missing something!

### DIFF
--- a/smpclient/__init__.py
+++ b/smpclient/__init__.py
@@ -65,7 +65,7 @@ class SMPClient:
     Args:
         transport: the `SMPTransport` to use
         address: the address of the SMP server, see `smpclient.transport` for details
-        timeout: the default timeout in seconds for SMP requests
+        timeout_s: the default timeout in seconds for SMP requests
 
     Example:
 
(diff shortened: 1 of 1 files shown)


smpclient/__init__.py
```diff
@@ -86,25 +89,30 @@ async def main():
     ```
     """
 
-    def __init__(self, transport: SMPTransport, address: str):  # noqa: DOC301
+    def __init__(
+        self, transport: SMPTransport, address: str, timeout: float = 60.0
```
JPHutchins: I realize that we initially had a very long default timeout. But I think that this should be `2.500`s default. Because most devices are local via USB, serial, BLE, or local network, and are expected to respond quickly. Let's discuss.

smpclient/__init__.py
```diff
@@ -86,25 +89,30 @@ async def main():
...
+    ):  # noqa: DOC301
         self._transport: Final = transport
         self._address: Final = address
+        self._timeout = timeout
```
JPHutchins:
```suggestion
        self._timeout_s: Final = timeout_s
```

smpclient/__init__.py
```diff
@@ -500,18 +506,18 @@ def _maximize_file_upload_packet(self, request: FileUpload, data: bytes) -> File
             len=request.len,
         )
 
-    async def _initialize(self) -> None:
+    async def _initialize(self, timeout_s: float | None = None) -> None:
```
JPHutchins: Adding a timeout param here is a really good idea.

smpclient/__init__.py
```diff
@@ -500,18 +506,18 @@ def _maximize_file_upload_packet(self, request: FileUpload, data: bytes) -> File
...
-        except asyncio.TimeoutError:
+            mcumgr_parameters = await self.request(MCUMgrParametersRead(), timeout_s=timeout_s)
+            if success(mcumgr_parameters):
+                logger.debug(f"MCUMgr parameters: {mcumgr_parameters}")
+                self._transport.initialize(mcumgr_parameters.buf_size)
+            elif error(mcumgr_parameters):
+                logger.warning(f"Error reading MCUMgr parameters: {mcumgr_parameters}")
+            else:
+                assert_never(mcumgr_parameters)
+        except TimeoutError:
```
JPHutchins: This bit is perhaps worth a change in the future. Technically, if this times out, it may indicate a situation in which the transport, but not the server, is available. Potential opportunity for fail fast, but also opportunity for confusing error messages on initialization due to a server that's slow to respond.

JPHutchins (changes requested):
This is a really great improvement, thank you! I'm mostly suggesting adding the _s seconds unit for consistency. Functionally, this is a great upgrade!

We do have a discussion about improving the architecture. Presently, "transport" and "client" are sorta intertwined in ways that prevent some nice context management or more flexible usage in general - including the "global transport options" issue faced by smpmgr. Would love your feedback if you have time: https://github.com/intercreate/smpclient/pull/58

I bring this up because ownership of timeout raises questions about the transport/client relationship - is the timeout parameter of the transport (SED example you gave) or the protocol (the initial upload request taking longer, maybe)?

PR #99, opened 2026-02-07
Update typing using pyupgrade

gudnimg:
A follow-up on https://github.com/intercreate/smpclient/pull/86 I later learned there exists a Python package which does this work automatically :) https://github.com/asottile/pyupgrade It catches a lot more stuff.

`pyupgrade` sometimes also fixes things unrelated to annotations/typing. So I added them in a separate commit so it can be reviewed on its own.


JPHutchins (approved):
Fantastic! That tool will keep being useful as long as Python keeps improving the type system. I look forward to dropping 3.11 and "legacy" generic syntax.

PR #105, opened 2026-05-02
Windows bonded BLE devices, connect via MAC

intercreate-gab:
This PR resolves connection to bonded devices without advertising on windows.


JPHutchins (commented):
Thanks for looking into this! Before proceeding, I'd like for you to take a look at this Bleak feature that was merged yesterday to see if it accomplishes the same without any workarounds: https://github.com/hbldh/bleak/pull/1979

We'd need to figure out ordering in the connect method. If the `get_connected_devices()` method is fast, then we could always run it first. If not, we'd have to consider racing it again the `find_device_by_address()`.

To test, you probably need to point the bleak version lock at github rather than a released version.

PR #107, opened 2026-05-22
feat: add bumble BLE transport with pairing

### LINKED ISSUE #15: Add Bumble as BLE backend
https://google.github.io/bumble/index.html

Having multiple BLE backends available is comprised of a few steps:

1. Add HW unit tests that perform SMP BLE operations
2. Define the common protocol required for an SMP BLE transport
3. Implement the protocol with Bleak and confirm no regression
4. Implement the protocol with Bumble and add to the HW unit tests

HW that I currently have available for HW unit tests:

- nrf52dk 52832
- nrf52dk 52840
- adafruit feather nrf52840

I also have a Boradcom BCM2070 USB BLE dongle (Plugable brand) that would be a good candidate for testing Bumble.

Sadly HW tests cannot run in GitHub workflow because this is a public repo: https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/about-self-hosted-runners#self-hosted-runner-security

I'd love to find a way!

JPHutchins:
## Summary

Adds an optional `bumble` extra (`smpclient[bumble]`) providing `SMPBumbleTransport` — a BLE `SMPTransport` backed by Google's [bumble](https://github.com/google/bumble) stack driving an external HCI USB controller (e.g. an nRF52840 DK running the Zephyr `hci_usb` sample).  Alternative to the bleak-backed `SMPBLETransport` for when the OS BLE stack is unavailable or reproducible cross-platform behavior is desired via the *same* HCI dongle everywhere.

Validated end-to-end against a real device: scan → pair (PIN flow) → reconnect with proactive `connection.encrypt()` from stored LTK → SMP Echo round-trip.

## What's in the PR

**Transport core** (`src/smpclient/transport/bumble/__init__.py`)
- `SMPBumbleTransport` with a `Disconnected | Connecting | Connected | ConnectedBorrowed` sum-type state machine; per-step `try/except` in `_teardown()` is load-bearing (skipping a bumble cleanup step can hang process exit).
- `ConnectedProtocol(Protocol)` — structural type shared by `Connected` (owned) and `ConnectedBorrowed` (caller-owned LE link).  `send`/`receive`/`mtu` operate on it via a single `_require_connected()` guard; no procedural None-sentinel checks.
- `pair_on_connect: PairingDelegate | None` constructor arg — `connect()` pairs after LE-connect and before GATT discovery when no bond exists; when a bond is present, proactive `_encrypt_using_bond` runs instead.
- `use_connection(connection, *, peer=None)` — borrow an existing bumble `Connection` for SMP traffic without owning its lifecycle.  Caller manages connect/disconnect/encryption.
- `bumble_device()` async context manager owns HCI transport + device lifecycle.
- `pair_device()` one-shot bonding (connect → pair → disconnect, no GATT).
- Module-level `pair()` function shared by `pair_device()`, `SMPBumbleTransport.pair()`, and the `pair_on_connect` path.
- `SMPBumbleTransport.scan()` static method using `bumble_device()`.

**Keystore** (`src/smpclient/transport/bumble/keystore.py`)
- `KeystoreStrategy = Tempfile | Local | Custom | ExistingCustom | InMemory` sum type.
- `Custom(path)` auto-creates parent dirs; `ExistingCustom(path)` errors if the file is missing.
- `Tempfile`/`Local` filenames validated as bare names (reject path separators).

**Pairing** (`src/smpclient/transport/bumble/pairing.py`)
- `NoInputNoOutput` / `KeyboardOnly(pin_cb)` / `DisplayOnly(display_cb)` delegates.
- `PairingResult = PairingSucceeded | PairingAlreadyBonded | PairingTimedOut | PairingFailed`.
- `PairingFailureReason` enum incl. `NOT_FOUND` for scan misses.

**Scan** (`src/smpclient/transport/bumble/scan.py`)
- Async BLE scan with name filter, eager mode, and SMP-service marker.
- Cleanup factored into `_scanning()` async context manager.

**CLI** (`src/smpclient/transport/bumble/__main__.py`)
- `smpbumble` entry point: `scan`, `pair [--force]`, `echo`.  Also `python -m smpclient.transport.bumble`.
- Typed `_ScanArgs` / `_PairArgs` / `_EchoArgs` NamedTuples for dispatch (no bare `argparse.Namespace` hand-off).

**Tests** (`tests/test_smp_bumble_transport.py`)
- 39 tests covering: state machine transitions, send/receive notification queue, pair delegates + `PairingResult` sum type, keystore strategies (Tempfile/Local/Custom/ExistingCustom/InMemory + filename validation), GATT discovery helpers, module-level `pair()`, `pair_device()` end-to-end with mocked bumble stack, `bumble_device()` context manager, scan helper, and CLI dispatch (`_prompt_pin` / `_scan` / `_pair` / `_echo` handlers + argparse routing).  Bumble module at 91%+ coverage.

**CI** (`.github/workflows/test.yaml`)
- New `bumble-only` row in `transport-extras` matrix.  Each row asserts the bumble import passes/fails as expected for that extras combination.

**Cross-cutting**
- `SMP_SERVICE_UUID` and `SMP_CHARACTERISTIC_UUID` centralized in `smpclient.transport`; re-exported by `ble.py` and `bumble/__init__.py`.
- `assert_never` imported from `typing_extensions` everywhere (Python 3.10 compat).
- `ATT_WRITE_OVERHEAD` named constant replaces magic `- 3` in MTU math.

## Test plan

- [x] `uv run task all` — lint, typecheck, full test suite (254 passed, 14 skipped)
- [x] `uv run task matrix` — same across Python 3.10–3.14
- [x] `uv run task coverage` — total coverage ≥ 91% gate
- [x] `uv run smpbumble scan` — lists all advertising devices, marks SMP ones
- [x] `uv run smpbumble pair <addr> [--force]` — pairs a device, optionally wiping local bond first
- [x] `uv run smpbumble echo <addr> <msg>` — connects, auto-encrypts from stored LTK, SMP Echo round-trip

## Follow-up (not blocking this PR)

- Firmware-lookup PR: companion PyPI packages from `intercreate/zephyr-hci` selectable via per-board smpclient extras + `all_firmware` aggregate (separate PR coordinating with the build farm side).

FYI @raykamp @raykamp-tht


JPHutchins:
```
uv run smpbumble pair 4E:D0:35:AE:33:0F --force
2026-05-21 18:10:14,618 smpclient.transport.bumble INFO force=True; deleting local bond for FB:C2:95:7E:16:A9
Enter the 6-digit PIN shown on the device: 567970
2026-05-21 18:10:26,033 smpclient.transport.bumble INFO Pair succeeded; settling for 1.5s
Pairing succeeded (bonded=True)
```

```
uv run smpbumble echo 67:1D:1F:D5:3E:23 hellow
2026-05-21 18:14:25,901 smpclient.transport.bumble INFO Connecting to 67:1D:1F:D5:3E:23
2026-05-21 18:14:25,929 smpclient.transport.bumble INFO Bond exists for FB:C2:95:7E:16:A9; initiating encryption
2026-05-21 18:14:27,848 smpclient.transport.bumble INFO Requested MTU 247, negotiated 247
2026-05-21 18:14:27,848 smpclient.transport.bumble INFO Connected to 67:1D:1F:D5:3E:23, max_write=244
hellow
2026-05-21 18:14:28,118 smpclient.transport.bumble WARNING Peer disconnected: reason=0x16
2026-05-21 18:14:28,122 smpclient.transport.bumble INFO Disconnected
```

src/smpclient/transport/bumble/__main__.py
```diff
@@ -0,0 +1,159 @@
...
+from bumble.device import Device
+from bumble.transport import open_transport
+
+from smpclient import SMPClient
+from smpclient.generics import error, success
+from smpclient.requests.os_management import EchoWrite
+from smpclient.transport.bumble import (
+    DEFAULT_HOST_ADDRESS,
+    DEFAULT_HOST_NAME,
+    SMP_SERVICE_UUID,
```
JPHutchins: Needs to be centralized outside of bleak or bumble modules.

src/smpclient/transport/bumble/__main__.py
```diff
@@ -0,0 +1,159 @@
...
+
+async def _prompt_pin() -> int | None:
+    raw = (await asyncio.to_thread(input, "Enter the 6-digit PIN shown on the device: ")).strip()
+    if not raw.isdigit():
+        print(f"Invalid PIN {raw!r}; expected digits only", file=sys.stderr)
+        return None
+    return int(raw)
+
+
+async def _scan(args: argparse.Namespace) -> int:
```
JPHutchins: Feels like we may want a scan static method on the transport and rely on logging and stuff here. Since the libs needs scanning capability.

src/smpclient/transport/bumble/__main__.py
```diff
@@ -0,0 +1,159 @@
...
+            print(f"Pairing timed out after {elapsed_s:.1f}s", file=sys.stderr)
+            return 1
+        case PairingFailed(reason, detail):
+            print(f"Pairing failed: {reason.value}: {detail}", file=sys.stderr)
+            return 1
+        case _:
+            assert_never(result)
+
+
+async def _echo(args: argparse.Namespace) -> int:
```
JPHutchins: Int return, what in the world is going on here?

src/smpclient/transport/bumble/__main__.py
```diff
@@ -0,0 +1,159 @@
...
+        if success(response):
+            print(response.r)
+            return 0
+        if error(response):
+            print(f"SMP error: {response}", file=sys.stderr)
+            return 1
+        return 2
+
+
+def smpbumble() -> int:
```
JPHutchins, line 110: Dispatched args must have strong static typing.

src/smpclient/transport/bumble/keystore.py
JPHutchins, line 1: We also need to be able to pass a path to an existing key store. TBD if that should require it to exist first.

src/smpclient/transport/bumble/scan.py
```diff
@@ -0,0 +1,129 @@
...
+            else (existing.name if existing else None),
+            rssi=rssi if rssi is not None else (existing.rssi if existing else None),
+            has_smp_service=(existing.has_smp_service if existing else False) or has_smp,
+        )
+        seen[address] = result
+        if first_match is not None and name is not None and result.name == name:
+            first_match.set()
+
+    device.on("advertisement", on_advertisement)
+    try:
```
JPHutchins, line 154: The block below this is pretty hard to read.

src/smpclient/transport/bumble/__init__.py
```diff
@@ -0,0 +1,621 @@
...
+        self._state: _State = Disconnected()
+
+        self._notifications: asyncio.Queue[bytes | _DisconnectSentinel] = asyncio.Queue()
+        self._disconnected_event = asyncio.Event()
+        self._disconnected_event.set()
+
+        logger.debug(f"Initialized {self.__class__.__name__}(hci={hci!r})")
+
+    @override
+    async def connect(self, address: str, timeout_s: float) -> None:
```
JPHutchins, line 189: I like the idea of having an alternative to connect, wherein the caller just lets the smpclient borrow the connection, and smpclient doesn't need to do connect or cleanup. This would allow the caller to use the BLE connection for non-SMP stuff, then use SMP, then back, all without disconnecting. May need to consider a lock, but ideally it would be the caller's problem.

src/smpclient/transport/bumble/__init__.py
```diff
@@ -0,0 +1,621 @@
...
+                f"post-pair state: is_encrypted={connection.is_encrypted}, "
+                f"authenticated={connection.authenticated}"
+            ),
+        )
+
+    @override
+    @property
+    def mtu(self) -> int:
+        if not isinstance(self._state, Connected):
+            return 0
```
JPHutchins: Eh???? Probably raise.

src/smpclient/transport/bumble/__init__.py
```diff
@@ -0,0 +1,621 @@
...
+        # that discards async return values, so this must stay sync.
+        logger.debug(f"Received notification: {len(data)} bytes")
+        self._notifications.put_nowait(data)
+
+    def _on_disconnection(self, reason: int) -> None:
+        logger.warning(f"Peer disconnected: reason=0x{reason:02x}")
+        self._disconnected_event.set()
+        self._notifications.put_nowait(_DisconnectSentinel())
+
+    async def _encrypt_if_bonded(self, connection: Connection, device: Device) -> None:
```
JPHutchins: Debug logs preferable to this essay that no one will ever see.

src/smpclient/transport/bumble/__init__.py
```diff
@@ -0,0 +1,621 @@
...
+            logger.warning(f"connection.encrypt() failed: {e}; subscribe may fail too")
+
+    async def _on_security_request(self, auth_req: object) -> None:
+        # Bumble emits this when a bonded peripheral asks us to start
+        # encryption.  We auto-encrypt only when an LTK exists; we never
+        # auto-initiate a fresh pairing.
+        if not isinstance(self._state, Connected):
+            return
+        connection = self._state.connection
+        device = self._state.device
```
JPHutchins: Why aliased?

src/smpclient/transport/bumble/__init__.py
```diff
@@ -0,0 +1,621 @@
...
+        try:
+            await connection.encrypt()
+        except Exception as e:
+            logger.warning(f"connection.encrypt() failed in response to security request: {e}")
+
+    async def _next_chunk(self) -> bytes:
+        if self._disconnected_event.is_set():
+            raise SMPTransportDisconnected(f"{self.__class__.__name__} peer disconnected")
+        chunk = await self._notifications.get()
+        match chunk:
```
JPHutchins: Walrus

src/smpclient/transport/bumble/__init__.py
```diff
@@ -0,0 +1,621 @@
...
+            case _:
+                assert_never(chunk)
+
+    async def _resolve_target(self, device: Device, address: str, timeout_s: float) -> str:
+        if MAC_ADDRESS_PATTERN.match(address):
+            return address
+
+        logger.info(f"Scanning for device name {address!r} (eager, up to {timeout_s}s)")
+        results = await scan_for_devices(device, timeout_s, name=address, eager=True)
+        if not results:
```
JPHutchins: Walrus

src/smpclient/transport/bumble/__init__.py
```diff
@@ -0,0 +1,621 @@
...
+            raise SMPBumbleTransportDeviceNotFound(
+                f"No advertising device matched name {address!r} in {timeout_s}s"
+            )
+        return results[0].address
+
+    async def _negotiate_mtu(self, peer: Peer, connection: Connection) -> int:
+        try:
+            negotiated = await peer.request_mtu(self._preferred_mtu)
+            logger.info(f"Requested MTU {self._preferred_mtu}, negotiated {negotiated}")
+            return negotiated - 3
```
JPHutchins: Magic number

src/smpclient/transport/bumble/__init__.py
```diff
@@ -0,0 +1,621 @@
...
+        except Exception as e:
+            logger.warning(
+                f"MTU exchange failed (preferred {self._preferred_mtu}), "
+                f"falling back to ATT MTU {connection.att_mtu}: {e}"
+            )
+            return connection.att_mtu - 3
```
JPHutchins: Magic

src/smpclient/transport/bumble/__init__.py
```diff
@@ -0,0 +1,621 @@
...
+
+
+async def _teardown(s: Connecting | Connected) -> None:
+    # Every step is wrapped so the next still runs: skipping a cleanup step
+    # because an earlier one raised leaves bumble in a state where subsequent
+    # operations — including process exit — may hang indefinitely.
```
JPHutchins: Yeah debug logging preferred

src/smpclient/transport/bumble/__init__.py
```diff
@@ -0,0 +1,621 @@
...
+    if s.transport is not None:
+        try:
+            await s.transport.close()
+        except Exception as e:
+            logger.warning(f"transport.close failed: {e}")
+
+
+def _find_smp_characteristic(peer: Peer) -> CharacteristicProxy[bytes]:
+    services = peer.get_services_by_uuid(BumbleUUID(str(SMP_SERVICE_UUID)))
+    if not services:
```
JPHutchins: Walrus

src/smpclient/transport/bumble/__init__.py
```diff
@@ -0,0 +1,621 @@
...
+        try:
+            yield device
+        finally:
+            try:
+                await device.power_off()
+            except Exception as e:
+                logger.warning(f"device.power_off failed: {e}")
+
+
+async def pair_device(
```
JPHutchins: It feels like there are 3 pairing impls in this PR?

src/smpclient/transport/bumble/__init__.py
```diff
@@ -0,0 +1,621 @@
...
+    """
+    async with bumble_device(hci=hci, delegate=delegate, keystore=keystore) as device:
+        ks: Final = device.keystore
+        assert ks is not None  # bumble_device always sets it
+
+        if MAC_ADDRESS_PATTERN.match(address):
+            target = address
+        else:
+            hits = await scan_for_devices(device, scan_timeout_s, name=address, eager=True)
+            if not hits:
```
JPHutchins: Walrus or anonymous

JPHutchins (commented):
Seems a lot of repetition, unnecessary variable aliasing, not enough use of Final.

JPHutchins:
PR description is out of date.

src/smpclient/transport/bumble/__init__.py
```diff
@@ -0,0 +1,621 @@
...
+Protocol.
+"""
+
+import asyncio
+import logging
+import re
+from contextlib import asynccontextmanager
+from dataclasses import dataclass
+from typing import AsyncIterator, Final, NamedTuple, TypeAlias, assert_never
+from uuid import UUID
```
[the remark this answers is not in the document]
JPHutchins: Done in 647b1cf — `assert_never` is imported from `typing_extensions` now.

src/smpclient/transport/bumble/keystore.py
```diff
@@ -0,0 +1,72 @@
...
+`SMPBumbleTransport` accepts a `KeystoreStrategy` that describes *where* bond
+keys (LTKs, IRKs, CSRKs) should be stored.  Persistence choice is exposed as a
+sum type so callers exhaustively handle the available options.
+"""
+
+import os
+import tempfile
+from pathlib import Path
+from typing import Final, NamedTuple, TypeAlias, assert_never
+
```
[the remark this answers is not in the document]
JPHutchins: Done in 647b1cf — fixed in `keystore.py`.

src/smpclient/transport/bumble/__main__.py
```diff
@@ -0,0 +1,159 @@
+"""Minimal CLI for the bumble SMP transport.
+
+Run `smpbumble --help` or `python -m smpclient.transport.bumble --help`.
+"""
+
+import argparse
+import asyncio
+import logging
+import sys
+from typing import assert_never
```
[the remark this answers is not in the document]
JPHutchins: Done in 647b1cf — fixed in `__main__.py`.

src/smpclient/transport/bumble/__init__.py
```diff
@@ -0,0 +1,621 @@
...
+    def _on_notification(self, data: bytes) -> None:
+        # bumble's CharacteristicProxy.subscribe() wraps with a sync on_change
+        # that discards async return values, so this must stay sync.
+        logger.debug(f"Received notification: {len(data)} bytes")
+        self._notifications.put_nowait(data)
+
+    def _on_disconnection(self, reason: int) -> None:
+        logger.warning(f"Peer disconnected: reason=0x{reason:02x}")
+        self._disconnected_event.set()
+        self._notifications.put_nowait(_DisconnectSentinel())
```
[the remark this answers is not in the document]
JPHutchins: Intentional — keeping. `_on_disconnection` sets `_disconnected_event` and pushes a `_DisconnectSentinel` onto `_notifications`, which makes any pending `send()`/`receive()` raise `SMPTransportDisconnected` per the `SMPTransport` Protocol contract. We deliberately avoid fire-and-forget `asyncio.create_task(self.disconnect())` — bumble's teardown is hang-prone and must run under explicit caller control. State stays `Connected` until the caller calls `disconnect()`, which is the canonical place to free bumble resources.

src/smpclient/transport/bumble/__init__.py
```diff
@@ -0,0 +1,621 @@
...
+            )
+        connection = self._state.connection
+        device = self._state.device
+
+        if device.keystore is not None:
+            existing = await device.keystore.get(str(connection.peer_address))
+            if existing is not None:
+                logger.info(f"Already bonded to {connection.peer_address}")
+                return PairingAlreadyBonded()
+
```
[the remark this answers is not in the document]
JPHutchins: Done in 647b1cf — both `SMPBumbleTransport.pair()` and the module-level `pair()` now take `force: bool = False`. When `force=False` and a bond exists, returns `PairingAlreadyBonded`. When `force=True`, the local bond is wiped and a fresh pairing is attempted. The CLI exposes this as `smpbumble pair <addr> --force`.

src/smpclient/transport/bumble/__init__.py
```diff
@@ -0,0 +1,621 @@
...
+    scan_timeout_s: float = 10.0,
+    pair_timeout_s: float = 30.0,
+    settle_s: float = DEFAULT_POST_PAIR_SETTLE_S,
+    force: bool = False,
+) -> PairingResult:
+    """One-shot bonding: open HCI, connect, pair, disconnect.
+
+    No GATT discovery or subscribe — use this to bond a peer before any SMP
+    traffic.  For peers whose SMP characteristic requires encryption, follow
+    up with `SMPBumbleTransport(pair_on_connect=delegate)`.
```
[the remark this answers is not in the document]
JPHutchins: Done in 647b1cf — `pair_on_connect: PairingDelegate | None = None` is now a real constructor arg on `SMPBumbleTransport`. When set, `connect()` pairs after LE-connect and before GATT discovery if no bond exists.

src/smpclient/transport/bumble/__init__.py
```diff
@@ -0,0 +1,621 @@
...
+
+        if MAC_ADDRESS_PATTERN.match(address):
+            target = address
+        else:
+            hits = await scan_for_devices(device, scan_timeout_s, name=address, eager=True)
+            if not hits:
+                return PairingFailed(
+                    reason=PairingFailureReason.USER_REJECTED,
+                    detail=f"no device matched name {address!r} within {scan_timeout_s}s",
+                )
```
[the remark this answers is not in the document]
JPHutchins: Done in 647b1cf — added `PairingFailureReason.NOT_FOUND` and use it for scan-by-name misses. `USER_REJECTED` is reserved for actual user rejection.

src/smpclient/transport/bumble/keystore.py
```diff
@@ -0,0 +1,72 @@
...
+    Returns:
+        A bumble `KeyStore` ready to be assigned to `Device.keystore`.
+    """
+    match strategy:
+        case Tempfile(filename):
+            return JsonKeyStore(
+                namespace=namespace,
+                filename=os.path.join(tempfile.gettempdir(), filename),
+            )
```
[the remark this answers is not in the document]
JPHutchins: Done in 647b1cf — added `_check_bare_filename()` that rejects absolute paths and any path separator (`os.sep`/`os.altsep`) on `Tempfile.filename` and `Local.filename`. Per @JPHutchins's adjacent comment, also added an `ExistingCustom(path: Path)` variant for an already-existing keystore file (raises `FileNotFoundError` if missing).

src/smpclient/transport/bumble/__main__.py
```diff
@@ -0,0 +1,159 @@
...
+    PairingTimedOut,
+)
+from smpclient.transport.bumble.scan import scan
+
+
+async def _prompt_pin() -> int | None:
+    raw = (await asyncio.to_thread(input, "Enter the 6-digit PIN shown on the device: ")).strip()
+    if not raw.isdigit():
+        print(f"Invalid PIN {raw!r}; expected digits only", file=sys.stderr)
+        return None
```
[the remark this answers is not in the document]
JPHutchins: Done in 647b1cf — `_prompt_pin` now requires exactly 6 digits and prints a clear error otherwise.

src/smpclient/transport/bumble/__init__.py
```diff
@@ -0,0 +1,621 @@
...
+    def __init__(
+        self,
+        *,
+        hci: str = DEFAULT_HCI_TRANSPORT,
+        host_address: Address = DEFAULT_HOST_ADDRESS,
+        host_name: str = DEFAULT_HOST_NAME,
+        keystore: KeystoreStrategy = Tempfile(),
+        preferred_mtu: int = DEFAULT_PREFERRED_MTU,
+    ) -> None:
+        """Initialize the bumble transport.
```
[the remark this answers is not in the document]
JPHutchins: Acknowledged — tests are explicitly deferred to a follow-up PR (called out in the updated PR description). Coverage drops below the 91% gate temporarily; will be restored when the test PR lands.

src/smpclient/transport/bumble/__main__.py
JPHutchins: Done in 647b1cf — `SMP_SERVICE_UUID` and `SMP_CHARACTERISTIC_UUID` now live in `smpclient.transport` (`src/smpclient/transport/__init__.py`). `ble.py` and `bumble/__init__.py` import from there.

src/smpclient/transport/bumble/__main__.py
JPHutchins: Done in 647b1cf — added `SMPBumbleTransport.scan(*, hci, timeout_s, service_uuid, name, eager)` static method using `bumble_device()` for the HCI lifecycle. CLI `_scan` calls it and just prints results.

src/smpclient/transport/bumble/__main__.py
JPHutchins: Done in 647b1cf — replaced the bare `return 2` with `assert_never(response)`. `success`/`error` are exhaustive TypeGuards so the branch is only reachable if a new response variant slips in, and mypy now catches that.

src/smpclient/transport/bumble/__main__.py
JPHutchins: Done in 647b1cf — added typed `_ScanArgs` / `_PairArgs` / `_EchoArgs` NamedTuples. The `smpbumble()` dispatcher builds the right NamedTuple per subcommand and passes it to a typed handler, instead of forwarding the bare `argparse.Namespace`.

src/smpclient/transport/bumble/keystore.py
JPHutchins: Done in 647b1cf — added `ExistingCustom(path: Path)` variant alongside `Custom(path: Path)`. `Custom` auto-creates the parent dir (existing behavior); `ExistingCustom` raises `FileNotFoundError` if the file isn't already there.

src/smpclient/transport/bumble/scan.py
JPHutchins: Done in 647b1cf — extracted the nested `try/finally` into a `_scanning()` async context manager. The main `scan()` body is a clean linear flow now.

src/smpclient/transport/bumble/__init__.py
JPHutchins: Deferred to follow-up — captured under "Deferred to follow-up PRs" in the updated PR description. The shape I have in mind: a `SMPBumbleTransport.from_existing_connection(connection, peer)` alt-ctor (or constructor flag) that skips `open_transport`/`device.connect`/`power_off`/`transport.close` so the caller owns the LE link.

src/smpclient/transport/bumble/__init__.py
JPHutchins: Done in 647b1cf — `mtu` now raises `SMPBumbleTransportException` when called outside the `Connected` state.

src/smpclient/transport/bumble/__init__.py
JPHutchins: Done in 647b1cf — trimmed `_encrypt_if_bonded`'s multi-line comment to a single line (and rolled the function into `_encrypt_using_bond` at module level, shared with `_on_security_request`).

src/smpclient/transport/bumble/__init__.py
JPHutchins: Done in 647b1cf — removed the `connection = self._state.connection` / `device = self._state.device` aliases. `SMPBumbleTransport.pair()` now delegates to module-level `pair(connection, device, ...)` directly.

src/smpclient/transport/bumble/__init__.py
JPHutchins: Done in 647b1cf — walrus in `_resolve_target`: `if not (results := await scan_for_devices(...)):`.

src/smpclient/transport/bumble/__init__.py
JPHutchins: Same as above — covered by the `_resolve_target` walrus in 647b1cf.

src/smpclient/transport/bumble/__init__.py
JPHutchins: Done in 647b1cf — named the magic `- 3` as `ATT_WRITE_OVERHEAD: Final = 3` (1-byte opcode + 2-byte handle = ATT write PDU header).

src/smpclient/transport/bumble/__init__.py
JPHutchins: Same — `ATT_WRITE_OVERHEAD` used in both branches of `_negotiate_mtu`.

src/smpclient/transport/bumble/__init__.py
JPHutchins: Done in 647b1cf — trimmed `_teardown`'s essay comment to a one-liner.

src/smpclient/transport/bumble/__init__.py
JPHutchins: Done in 647b1cf — walrus in `_find_smp_characteristic`: `if not (services := peer.get_services_by_uuid(...)):` and similarly for `characteristics`.

src/smpclient/transport/bumble/__init__.py
JPHutchins: Addressed in 647b1cf — there's now one module-level `pair()` shared by `SMPBumbleTransport.pair()`, `pair_device()`, and the `pair_on_connect` path inside `connect()`. Encryption-with-existing-LTK is `_encrypt_using_bond()` and is shared between proactive (`connect()`) and reactive (`_on_security_request`) paths. Down to one pairing function and one encryption function.

src/smpclient/transport/bumble/__init__.py
JPHutchins: Done in 647b1cf — walrus in `pair_device`: `elif hits := await scan_for_devices(device, scan_timeout_s, name=address, eager=True):`.

src/smpclient/transport/bumble/__init__.py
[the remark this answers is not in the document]
JPHutchins: Update: unit tests are now in this PR — `tests/test_smp_bumble_transport.py` has 39 tests covering the state machine, send/receive notify queue, pair delegates + result sum type, all five keystore strategies + filename validation, GATT discovery helpers, the module-level `pair()`, `pair_device()` end-to-end with a mocked bumble stack, `bumble_device()` ctx manager, scan, and CLI dispatch.  Bumble module coverage is now 91%+ and the overall coverage gate (91%) passes.  Shipped in 6d7e25d.

src/smpclient/transport/bumble/__init__.py
JPHutchins: Update: this is now in the PR (6d7e25d).  `SMPBumbleTransport.use_connection(connection, *, peer=None)` lets a caller hand the transport an existing bumble `Connection` — we do GATT discovery + subscribe on it, transition to a new `ConnectedBorrowed` state, and on `disconnect()` we only unsubscribe; the caller keeps ownership of the LE link, the `Device`, and the HCI transport.  `Connected` and `ConnectedBorrowed` share a `ConnectedProtocol` structural type so `send`/`receive`/`mtu` work on either without procedural sentinel checks.

src/smpclient/transport/bumble/keystore.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+def _check_bare_filename(filename: str) -> None:
+    # `Tempfile` and `Local` join `filename` onto a system-chosen directory.
+    # Absolute paths or path separators would let the caller escape that
+    # directory; reject them so the strategy contract holds.
+    if os.path.isabs(filename) or os.sep in filename or (os.altsep and os.altsep in filename):
+        raise InvalidKeystoreFilename(
+            f"{filename!r} must be a bare filename — use Custom(path=...) for an arbitrary location"
+        )
```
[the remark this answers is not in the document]
JPHutchins: Done in 9a7073d — `_check_bare_filename` now uses `pathlib.PurePath`: it requires `p.name == filename` and rejects empty strings, `.`/`..`, and Windows drive prefixes like `C:foo` (`p.drive == p.root == ''`).

src/smpclient/transport/bumble/keystore.py
```diff
@@ -0,0 +1,102 @@
...
+            data_dir: Final = Path(user_data_dir("smpclient", "intercreate"))
+            data_dir.mkdir(parents=True, exist_ok=True)
+            return JsonKeyStore(namespace=namespace, filename=str(data_dir / filename))
+        case Custom(path):
+            path.parent.mkdir(parents=True, exist_ok=True)
+            return JsonKeyStore(namespace=namespace, filename=str(path))
+        case ExistingCustom(path):
+            if not path.exists():
+                raise FileNotFoundError(f"ExistingCustom keystore not found: {path}")
+            return JsonKeyStore(namespace=namespace, filename=str(path))
```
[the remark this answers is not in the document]
JPHutchins: Done in 9a7073d — `ExistingCustom` checks `path.is_file()` (not just `path.exists()`).  A directory, broken symlink, or special file now raises a clear `FileNotFoundError` at `resolve()` time with a message that names the path.

src/smpclient/transport/bumble/__init__.py
```diff
@@ -0,0 +1,710 @@
...
+        keystore = self._standalone_keystore()
+        if address is None:
+            await keystore.delete_all()
+            logger.info("Cleared all bonds")
+        else:
+            await keystore.delete(address)
+            logger.info(f"Cleared bond for {address}")
+
+    def _standalone_keystore(self) -> KeyStore:
+        return resolve_keystore(self._keystore, namespace=str(self._host_address))
```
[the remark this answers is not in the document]
JPHutchins: Done in 9a7073d — `connect()` and `bumble_device()` both pass `namespace=str(host_address)` now, the same value `_standalone_keystore` uses.  `bonded_devices()`/`clear_bond()` see the same namespace `connect()` writes into.

src/smpclient/transport/bumble/scan.py
```diff
@@ -0,0 +1,141 @@
...
+            matching results are returned.
+        eager: When True with a `name` filter, returns at the first match.
+
+    Returns:
+        Observed `ScanResult`s, deduplicated by address.
+    """
+    target: Final = BumbleUUID(str(service_uuid)) if service_uuid is not None else None
+    seen: Final[dict[str, ScanResult]] = {}
+    first_match = asyncio.Event() if eager else None
+
```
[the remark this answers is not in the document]
JPHutchins: Done in 9a7073d — replaced the `name`/`eager` pair with a `ScanMode` sum type: `ScanAll()` vs `ScanForName(name, eager=True)`.  Invalid states (eager without a name) are now unrepresentable at the type level.  `ScanForName` keeps `eager` as a flag (default True) per a user follow-up, so callers can enumerate multiple peers sharing a name with `ScanForName(name, eager=False)`.

src/smpclient/transport/bumble/__init__.py
```diff
@@ -0,0 +1,710 @@
...
+
+        Args:
+            hci: bumble HCI transport spec.
+            timeout_s: Maximum scan duration.
+            service_uuid: When set, results with this UUID get
+                `has_smp_service=True`.  Defaults to the SMP service UUID.
+            name: When set together with `eager=True`, the scan returns at
+                the first match.
+            eager: Return at first match.  Requires `name`.
+
```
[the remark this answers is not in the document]
JPHutchins: Done in 9a7073d — `SMPBumbleTransport.scan()` now takes `mode: ScanMode = ScanAll()` instead of `name` + `eager`, so the invariant is enforced by the type, not the docstring.

src/smpclient/transport/firmware/hci.py
```diff
@@ -0,0 +1,22 @@
...
+    PosixPath('.../zephyr_4_4_0_hci_usb_nrf52840dk_default/firmware.hex')
+"""
+
+try:
+    from zephyr_4_4_0_hci import Firmware, firmware
+except ModuleNotFoundError as e:
+    raise ModuleNotFoundError(
+        "Bundled Zephyr HCI firmware is not installed.  Install with "
+        "`pip install smpclient[hci_firmware]` (or `pip install zephyr-4.4.0-hci`)."
+    ) from e
```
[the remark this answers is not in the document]
JPHutchins: Done in 73f933c — also moved to `smpclient.transport.firmware.hci` so transitive deps imported by `zephyr_4_4_0_hci` would attach `name=` of the missing transitive module, not `zephyr_4_4_0_hci`. The narrow `if e.name != "zephyr_4_4_0_hci": raise` lets that propagate.  Covered by `test_unrelated_module_not_found_is_reraised`.

src/smpclient/transport/bumble/scan.py
```diff
@@ -0,0 +1,167 @@
...
+    Args:
+        device: A bumble `Device` already powered on.
+        timeout_s: Maximum scan duration in seconds.
+        mode: A `ScanMode` sum type.  `ScanAll()` returns everything observed
+            for the full timeout.  `ScanForName(name, eager=True)` (default)
+            returns at the first advertisement matching `name`;
+            `ScanForName(name, eager=False)` runs the full timeout and returns
+            every matching device — useful for enumerating peers that share
+            a name.
```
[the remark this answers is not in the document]
JPHutchins: Done in 73f933c — docstring now correctly reports `ScanAll()` as the default.

src/smpclient/transport/bumble/__init__.py
```diff
@@ -0,0 +1,802 @@
...
+        detail=(
+            f"post-pair state: is_encrypted={connection.is_encrypted}, "
+            f"authenticated={connection.authenticated}"
+        ),
+    )
+
+
+async def _teardown_borrowed(s: ConnectedBorrowed) -> None:
+    # Borrowed: caller owns the LE link, device, and HCI transport.  We only
+    # unsubscribe and drop our listener.
```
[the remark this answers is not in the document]
JPHutchins: Done in 73f933c — `_teardown_borrowed` now also calls `connection.remove_listener(EVENT_DISCONNECTION, on_disconnection)` so the caller's Connection isn't left holding references to the transport after we hand it back.

src/smpclient/transport/bumble/__init__.py
```diff
@@ -0,0 +1,802 @@
...
+        *,
+        peer: Peer | None = None,
+    ) -> None:
+        """Use an existing bumble `Connection` for SMP traffic; caller owns the LE link.
+
+        The transport does GATT discovery + subscribe on the supplied connection
+        and transitions to `Connected`.  On `disconnect()`, the transport
+        unsubscribes but does *not* tear down the connection, device, or HCI
+        transport — the caller is responsible for those.  This lets a single
+        bumble session multiplex SMP and non-SMP traffic over the same link.
```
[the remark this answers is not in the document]
JPHutchins: Done in 73f933c — docstring corrected to `ConnectedBorrowed`.

src/smpclient/transport/bumble/__init__.py
```diff
@@ -0,0 +1,802 @@
...
+It is an alternative to `smpclient.transport.ble.SMPBLETransport`, which uses
+the OS's BLE stack via bleak.  The bumble transport is useful when the OS
+stack is unavailable, missing required features (e.g. LE Secure Connections
+with a custom IO capability), or when reproducible cross-platform behavior is
+desired by using the *same* HCI controller everywhere.
+
+The transport's connection lifecycle is modeled as a sum type
+(`Disconnected | Connecting | Connected`).  Pattern-match on `transport._state`
+or, better, treat the transport as a black box through the `SMPTransport`
+Protocol.
```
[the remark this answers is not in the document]
JPHutchins: Done in 73f933c — module docstring now lists `Disconnected | Connecting | Connected | ConnectedBorrowed` and explains the owned-vs-borrowed split.

src/smpclient/transport/bumble/__init__.py
```diff
@@ -0,0 +1,811 @@
...
+        logger.info(f"Borrowing connection to {connection.peer_address}, max_write={max_write}")
+
+    async def bonded_devices(self) -> tuple[str, ...]:
+        """Return the BD_ADDRs of peers currently in the keystore.
+
+        Works in any state — the keystore is constructed from the configured
+        strategy + the local host address.
+        """
+        bonds = await self._standalone_keystore().get_all()
+        return tuple(addr for addr, _keys in bonds)
```
JPHutchins: inline `bonds`

src/smpclient/transport/bumble/__init__.py
```diff
@@ -0,0 +1,811 @@
...
+            smp_characteristic=smp_char,
+            max_write=max_write,
+        )
+        logger.info(f"Borrowing connection to {connection.peer_address}, max_write={max_write}")
+
+    async def bonded_devices(self) -> tuple[str, ...]:
+        """Return the BD_ADDRs of peers currently in the keystore.
+
+        Works in any state — the keystore is constructed from the configured
+        strategy + the local host address.
```
JPHutchins: Remove unnecessary impl detail.

src/smpclient/transport/bumble/__init__.py
```diff
@@ -0,0 +1,811 @@
...
+            address: The peer BD_ADDR whose bond should be removed.  If
+                `None`, every bond in the keystore is removed.
+        """
+        keystore = self._standalone_keystore()
+        if address is None:
+            await keystore.delete_all()
+            logger.info("Cleared all bonds")
+        else:
+            await keystore.delete(address)
+            logger.info(f"Cleared bond for {address}")
```
JPHutchins: Split into two functions without a None coded meaning: `clear_bond` and `clear_bonds`

src/smpclient/transport/bumble/__init__.py
```diff
@@ -0,0 +1,811 @@
...
+    def _on_disconnection(self, reason: int) -> None:
+        logger.warning(f"Peer disconnected: reason=0x{reason:02x}")
+        self._disconnected_event.set()
+        self._notifications.put_nowait(_DisconnectSentinel())
+
+    async def _on_security_request(self, auth_req: object) -> None:
+        # Auto-encrypt only when an LTK exists; never auto-initiate a fresh pair.
+        # `ConnectedBorrowed` skips this: the caller owns the Device & its keystore.
+        if not isinstance(self._state, Connected):
+            return
```
JPHutchins: Let's info or warn here? I feel that ConnectBorrowed would be logged differently than an unexpected state too.

src/smpclient/transport/bumble/__init__.py
```diff
@@ -0,0 +1,811 @@
...
+        logger.debug(f"Peer security request: {auth_req!r}")
+        await _encrypt_using_bond(self._state.connection, self._state.device)
+
+    async def _next_chunk(self) -> bytes:
+        if self._disconnected_event.is_set():
+            raise SMPTransportDisconnected(f"{self.__class__.__name__} peer disconnected")
+        chunk = await self._notifications.get()
+        match chunk:
```
JPHutchins: walrus

src/smpclient/transport/bumble/__init__.py
```diff
@@ -0,0 +1,811 @@
...
+            return negotiated - ATT_WRITE_OVERHEAD
+        except Exception as e:
+            logger.warning(
+                f"MTU exchange failed (preferred {self._preferred_mtu}), "
+                f"falling back to ATT MTU {connection.att_mtu}: {e}"
+            )
+            return connection.att_mtu - ATT_WRITE_OVERHEAD
+
+
+async def _encrypt_using_bond(connection: Connection, device: Device) -> bool:
```
JPHutchins: Move out of __init__.py

JPHutchins: It also seems to be doing a lot without logging or giving the caller context on the failure. I'd think that returning an enumeration of outcomes would be most appropriate here so that the caller can decide what action to take.

src/smpclient/transport/bumble/__init__.py
```diff
@@ -0,0 +1,811 @@
...
+    logger.debug(f"Bond exists for {connection.peer_address}; initiating encryption")
+    try:
+        await connection.encrypt()
+    except Exception as e:
+        logger.warning(f"connection.encrypt() failed: {e}")
+        return False
+    return connection.is_encrypted
+
+
+async def pair(
```
JPHutchins: Move out of __init__.py, applies to generic bumble not just an smpclient transport

src/smpclient/transport/bumble/__init__.py
```diff
@@ -0,0 +1,811 @@
...
+        )
+    ):
+        raise SMPBumbleTransportNotSMPServer(
+            f"SMP characteristic {SMP_CHARACTERISTIC_UUID} not found on peer"
+        )
+    return characteristics[0]
+
+
+@asynccontextmanager
+async def bumble_device(
```
JPHutchins: Move out of __init__.py, applies to generic bumble not just an smpclient transport

src/smpclient/transport/bumble/__init__.py
```diff
@@ -0,0 +1,811 @@
...
+        try:
+            yield device
+        finally:
+            try:
+                await device.power_off()
+            except Exception as e:
+                logger.warning(f"device.power_off failed: {e}")
+
+
+async def pair_device(
```
JPHutchins: Move out of __init__.py, applies to generic bumble not just an smpclient transport

src/smpclient/transport/bumble/__init__.py
```diff
@@ -0,0 +1,811 @@
...
+            service_uuid: When set, results advertising this UUID get
+                `has_smp_service=True`.  Defaults to the SMP service UUID.
+
+        Returns:
+            Observed `ScanResult`s.
+        """
+        async with bumble_device(hci=hci) as device:
+            return await scan_for_devices(device, timeout_s, mode, service_uuid=service_uuid)
+
+    async def use_connection(
```
JPHutchins, line 334: Can we also add an async context manager for this (at bottom of file as a function)?

JPHutchins: Actually, then this would use that async context manager internally.

src/smpclient/transport/bumble/__init__.py
```diff
@@ -0,0 +1,811 @@
...
+        self._preferred_mtu: Final = preferred_mtu
+        self._pair_on_connect: Final = pair_on_connect
+        self._pair_timeout_s: Final = pair_timeout_s
+        self._settle_s: Final = settle_s
+
+        self._state: _State = Disconnected()
+
+        self._notifications: asyncio.Queue[bytes | _DisconnectSentinel] = asyncio.Queue()
+        self._disconnected_event = asyncio.Event()
+        self._disconnected_event.set()
```
JPHutchins: These are all Final

src/smpclient/transport/bumble/__init__.py
JPHutchins, line 1: Generally this file is too long. Comments can be trimmed down somewhat. Any comment that explains what the code is doing or how it is doing it should be removed. The code already explains what it's doing, and if it doesn't, then it's bad code.

Otherwise there are various suggestions throughout.

JPHutchins (commented):
Review round focused on cleaning up the new transport's` __init__.py`

src/smpclient/transport/bumble/__init__.py
JPHutchins: Done in 75b1e4e — moved `bumble_device()`, `pair()`, `pair_device()`, and `encrypt_using_bond()` (plus their constants) out of `__init__.py` into `bumble/device.py` and `bumble/pairing.py`.  `__init__.py` is now ~290 lines shorter (mostly the four moved functions) and the WHAT/HOW narrative comments (e.g. the `_teardown_borrowed` block) are gone.  Comments that remain are WHY (sync-vs-async constraint, per-step try/except being load-bearing, LTK-only auto-encrypt rule).

src/smpclient/transport/bumble/__init__.py
JPHutchins: Done in 75b1e4e — `self._notifications` and `self._disconnected_event` now have `Final` annotations.  `self._state` stays mutable (reassigned on transitions).

src/smpclient/transport/bumble/__init__.py
JPHutchins: Done in 75b1e4e — `borrowed_connection(transport, connection, *, peer)` is at the bottom of `__init__.py`; wraps `use_connection()` + `disconnect()` so callers can `async with borrowed_connection(...):`.

src/smpclient/transport/bumble/__init__.py
JPHutchins: Done in 75b1e4e — moved to `smpclient.transport.bumble.pairing`.

src/smpclient/transport/bumble/__init__.py
JPHutchins: Done in 75b1e4e — moved to `smpclient.transport.bumble.device` alongside the host/HCI constants it uses.

src/smpclient/transport/bumble/__init__.py
JPHutchins: Done in 75b1e4e — moved to `smpclient.transport.bumble.pairing` (natural home — `pairing.py` already held the delegates + result types).

src/smpclient/transport/bumble/__init__.py
JPHutchins: Done in 75b1e4e — moved to `smpclient.transport.bumble.pairing` as the public `encrypt_using_bond()`.  It's pairing-adjacent (consults the keystore for a stored LTK).

src/smpclient/transport/bumble/__init__.py
JPHutchins: Done in 75b1e4e — `match (chunk := await self._notifications.get()):`.

src/smpclient/transport/bumble/__init__.py
JPHutchins: Done in 75b1e4e — replaced the early-return with a `match self._state`: `Connected` → debug + encrypt-using-bond, `ConnectedBorrowed` → info (caller owns the Device & keystore), `Disconnected | Connecting` → warning (unexpected).

src/smpclient/transport/bumble/__init__.py
JPHutchins: Done in 75b1e4e — `clear_bond(address: str)` and `clear_bonds()` are now separate methods, no `None`-coded meaning.

src/smpclient/transport/bumble/__init__.py
JPHutchins: Done in 75b1e4e — dropped the 'Works in any state — the keystore is constructed from the configured strategy + the local host address' paragraph.

src/smpclient/transport/bumble/__init__.py
JPHutchins: Done in 75b1e4e — single expression: `return tuple(addr for addr, _keys in await self._standalone_keystore().get_all())`.

src/smpclient/transport/bumble/__init__.py
```diff
@@ -0,0 +1,630 @@
...
+desired by using the *same* HCI controller everywhere.
+
+The transport's connection lifecycle is modeled as a sum type
+(`Disconnected | Connecting | Connected | ConnectedBorrowed`).  `Connected`
+is for connections owned by the transport (opened via `connect()`);
+`ConnectedBorrowed` is for caller-owned connections adopted via
+`use_connection()` and is torn down without closing the underlying transport.
+Pattern-match on `transport._state` or, better, treat the transport as a
+black box through the `SMPTransport` Protocol.
+"""
```
JPHutchins: Remove most of this docstring as it duplicates what is plainly obvious from the source code violating SSOT and increasing maintenance burden.

src/smpclient/transport/bumble/__init__.py
```diff
@@ -0,0 +1,630 @@
...
+    PairingResult,
+    PairingSucceeded,
+    PairingTimedOut,
+    encrypt_using_bond,
+    pair,
+)
+from smpclient.transport.bumble.scan import ScanAll, ScanForName, ScanMode, ScanResult
+from smpclient.transport.bumble.scan import scan as scan_for_devices
+
+logger = logging.getLogger(__name__)
```
JPHutchins: Final

src/smpclient/transport/bumble/__init__.py
```diff
@@ -0,0 +1,630 @@
...
+
+class SMPBumbleTransportNotSMPServer(SMPBumbleTransportException):
+    """Raised when the connected peripheral does not expose the SMP characteristic."""
+
+
+class Disconnected(NamedTuple): ...
+
+
+class _DisconnectSentinel(NamedTuple):
+    """Pushed onto `_notifications` to wake a pending `receive()` on disconnect."""
```
JPHutchins: Remove impl detail from docstring

src/smpclient/transport/bumble/__init__.py
```diff
@@ -0,0 +1,630 @@
...
+
+
+@dataclass(slots=True)
+class Connecting:
+    """Partial state populated field-by-field during `connect()`, drained by `disconnect()` on failure."""
```
JPHutchins: Remove impl detail from docstring

src/smpclient/transport/bumble/__init__.py
```diff
@@ -0,0 +1,630 @@
...
+    connection: Connection | None = None
+    peer: Peer | None = None
+    smp_characteristic: CharacteristicProxy[bytes] | None = None
+
+
+class ConnectedProtocol(Protocol):
+    """Structural type for any state with an established SMP channel.
+
+    Both `Connected` (owned) and `ConnectedBorrowed` (caller-owned LE link)
+    satisfy this — `send`/`receive`/`mtu` operate on it without caring which.
```
JPHutchins: How can the protocol know what satisfies it? Remove this noise.

src/smpclient/transport/bumble/__init__.py
```diff
@@ -0,0 +1,630 @@
...
+    @property
+    def peer(self) -> Peer: ...
+    @property
+    def smp_characteristic(self) -> CharacteristicProxy[bytes]: ...
+    @property
+    def max_write(self) -> int: ...
+
+
+class Connected(NamedTuple):
+    """Owned connection: transport built the link and tears it all down on disconnect."""
```
JPHutchins: Remover impl detail from docstring

src/smpclient/transport/bumble/__init__.py
```diff
@@ -0,0 +1,630 @@
...
+    transport: Transport
+    device: Device
+    connection: Connection
+    peer: Peer
+    smp_characteristic: CharacteristicProxy[bytes]
+    max_write: int
+
+
+class ConnectedBorrowed(NamedTuple):
+    """Borrowed connection: caller owns transport/device and the LE link itself."""
```
JPHutchins: Remover impl detail from docstring

src/smpclient/transport/bumble/__init__.py
```diff
@@ -0,0 +1,630 @@
...
+                f"use_connection() called while in state {type(self._state).__name__}"
+            )
+
+        while not self._notifications.empty():
+            self._notifications.get_nowait()
+        self._disconnected_event.clear()
+
+        connection.on(Connection.EVENT_DISCONNECTION, self._on_disconnection)
+
+        p: Final = peer if peer is not None else Peer(connection)
```
JPHutchins: Walrus

src/smpclient/transport/bumble/__init__.py
```diff
@@ -0,0 +1,630 @@
...
+        # that discards async return values, so this must stay sync.
+        logger.debug(f"Received notification: {len(data)} bytes")
+        self._notifications.put_nowait(data)
+
+    def _on_disconnection(self, reason: int) -> None:
+        logger.warning(f"Peer disconnected: reason=0x{reason:02x}")
+        self._disconnected_event.set()
+        self._notifications.put_nowait(_DisconnectSentinel())
+
+    async def _on_security_request(self, auth_req: object) -> None:
```
JPHutchins: Why is object the best typing we can get here? Is bumble that poorly typed?

src/smpclient/transport/bumble/__init__.py
```diff
@@ -0,0 +1,630 @@
...
+    async def _resolve_target(self, device: Device, address: str, timeout_s: float) -> str:
+        if MAC_ADDRESS_PATTERN.match(address):
+            return address
+
+        logger.info(f"Scanning for device name {address!r} (eager, up to {timeout_s}s)")
+        if not (results := await scan_for_devices(device, timeout_s, ScanForName(address))):
+            raise SMPBumbleTransportDeviceNotFound(
+                f"No advertising device matched name {address!r} in {timeout_s}s"
+            )
+        return results[0].address
```
JPHutchins: This does not appear to use self, can it be moved to the generic helper files?

src/smpclient/transport/bumble/__init__.py
```diff
@@ -0,0 +1,630 @@
...
+        try:
+            negotiated: Final = await peer.request_mtu(self._preferred_mtu)
+            logger.info(f"Requested MTU {self._preferred_mtu}, negotiated {negotiated}")
+            return negotiated - ATT_WRITE_OVERHEAD
+        except Exception as e:
+            logger.warning(
+                f"MTU exchange failed (preferred {self._preferred_mtu}), "
+                f"falling back to ATT MTU {connection.att_mtu}: {e}"
+            )
+            return connection.att_mtu - ATT_WRITE_OVERHEAD
```
JPHutchins: Doesn't use self, can move out to helpers?

src/smpclient/transport/bumble/__init__.py
```diff
@@ -0,0 +1,630 @@
...
+
+async def _teardown_borrowed(s: ConnectedBorrowed, on_disconnection: Callable[[int], None]) -> None:
+    try:
+        await s.smp_characteristic.unsubscribe()
+    except Exception as e:
+        logger.warning(f"smp_characteristic.unsubscribe failed: {e}")
+    try:
+        s.connection.remove_listener(Connection.EVENT_DISCONNECTION, on_disconnection)
+    except Exception as e:
+        logger.warning(f"remove_listener(EVENT_DISCONNECTION) failed: {e}")
```
JPHutchins: May make sense to make a private method to avoid having to pass the callback (can just get everything from self and lower complexity)

src/smpclient/transport/bumble/pairing.py
```diff
@@ -0,0 +1,256 @@
+"""Pairing delegates, results, and module-level pairing helpers.
+
+Three IO-capability flavors are provided directly; users can subclass
+`bumble.pairing.PairingDelegate` for more exotic flows.
+
+Pairing outcomes are exposed as a sum type (`PairingResult`) so callers
+exhaustively handle every case.  See `SMPBumbleTransport.pair`.
```
JPHutchins: Remover impl detail from docstring

src/smpclient/transport/bumble/pairing.py
```diff
@@ -0,0 +1,256 @@
...
+from smpclient.transport.bumble.scan import scan as scan_for_devices
+
+logger = logging.getLogger(__name__)
+
+
+DEFAULT_PAIR_TIMEOUT_S: Final = 30.0
+DEFAULT_POST_PAIR_SETTLE_S: Final = 1.5
+"""After `connection.pair()` returns, peers need a brief settle window to
+finalize bonding on their side before disconnect.  Tune per-peer if a
+particular peer needs longer."""
```
JPHutchins: Remover impl detail from docstring

src/smpclient/transport/firmware/hci.py
```diff
@@ -0,0 +1,23 @@
...
+
+Importing this module requires the `smpclient[hci_firmware]` extra, which
+pulls in `zephyr-4.4.0-hci` — an umbrella that depends on every published
+`hci_usb` firmware variant.  The umbrella exposes `firmware: Firmware`, a
+NamedTuple whose fields are the per-variant modules, so callers get typed
+attribute access plus IDE autocomplete:
+
+    >>> from smpclient.transport.firmware.hci import firmware
+    >>> firmware.nrf52840dk_default.HEX_PATH  # doctest: +SKIP
+    PosixPath('.../zephyr_4_4_0_hci_usb_nrf52840dk_default/firmware.hex')
```
JPHutchins: Remover impl detail from docstring

JPHutchins (commented):
More cleanup.

src/smpclient/transport/bumble/__init__.py
JPHutchins: Done in 5f92a75 — module docstring is now one line.

src/smpclient/transport/bumble/__init__.py
JPHutchins: Done in 5f92a75 — `logger: Final = logging.getLogger(__name__)`.  Also applied to `pairing.py`.

src/smpclient/transport/bumble/__init__.py
JPHutchins: Done in 5f92a75 — `_DisconnectSentinel` is now `class _DisconnectSentinel(NamedTuple): ...` (no docstring).

src/smpclient/transport/bumble/__init__.py
JPHutchins: Done in 5f92a75 — `Connecting` docstring dropped.

src/smpclient/transport/bumble/__init__.py
JPHutchins: Done in 5f92a75 — dropped the noise about which states satisfy `ConnectedProtocol`.

src/smpclient/transport/bumble/__init__.py
JPHutchins: Done in 5f92a75 — `Connected` docstring dropped.

src/smpclient/transport/bumble/__init__.py
JPHutchins: Done in 5f92a75 — `ConnectedBorrowed` docstring dropped.

src/smpclient/transport/bumble/__init__.py
JPHutchins: Done in 5f92a75 — `if not (p := peer if peer is not None else Peer(connection)).services:`.

src/smpclient/transport/bumble/__init__.py
JPHutchins: Done in 5f92a75 — typed as `AuthReq` from `bumble.smp`.  bumble does export it; the `Connection.EVENT_SECURITY_REQUEST` is emitted with `request.auth_req: AuthReq` (a `SpecableFlag` enum).

src/smpclient/transport/bumble/__init__.py
JPHutchins: Done in 5f92a75 — `_resolve_target` is now a module-level function (no `self`).  Kept in `__init__.py` because it raises `SMPBumbleTransportDeviceNotFound` (avoids circular import with `device.py`).

src/smpclient/transport/bumble/__init__.py
JPHutchins: Done in 5f92a75 — `_negotiate_mtu(peer, connection, preferred_mtu)` is now module-level.  Kept in `__init__.py` next to `ATT_WRITE_OVERHEAD`.

src/smpclient/transport/bumble/__init__.py
JPHutchins: Done in 5f92a75 — `_teardown_borrowed` is back on the class as `self._teardown_borrowed()` with no args; it reads `self._state` (asserted `ConnectedBorrowed`) and `self._on_disconnection` directly.

src/smpclient/transport/bumble/pairing.py
JPHutchins: Done in 5f92a75 — pairing.py module docstring trimmed to one line.

src/smpclient/transport/bumble/pairing.py
JPHutchins: Done in 5f92a75 — trimmed the `DEFAULT_POST_PAIR_SETTLE_S` docstring and the `PinCallback`/`DisplayCallback`/`PairingSucceeded.bonded`/`PairingAlreadyBonded` docstrings.  Class docstrings trimmed to one line each.

src/smpclient/transport/firmware/hci.py
JPHutchins: Done in 5f92a75 — `firmware/hci.py` module docstring is now one line.

src/smpclient/transport/bumble/scan.py
```diff
@@ -0,0 +1,167 @@
...
+        advertisement.data.SHORTENED_LOCAL_NAME,
+    ):
+        value = advertisement.data.get(ad_type)
+        if isinstance(value, str):
+            return value
+        if isinstance(value, bytes):
+            try:
+                return value.decode("utf-8")
+            except UnicodeDecodeError:
+                return None
```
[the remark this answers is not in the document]
JPHutchins: Done in 5f92a75 — `_extract_name` now `continue`s on `UnicodeDecodeError` so a bad COMPLETE_LOCAL_NAME falls through to SHORTENED_LOCAL_NAME.

src/smpclient/transport/firmware/hci.py
```diff
@@ -0,0 +1,23 @@
...
+"""
+
+try:
+    from zephyr_4_4_0_hci import Firmware as Firmware
+    from zephyr_4_4_0_hci import firmware as firmware
+except ModuleNotFoundError as e:
+    if e.name != "zephyr_4_4_0_hci":
+        raise
+    raise ModuleNotFoundError(
```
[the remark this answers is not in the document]
JPHutchins: Done in 5f92a75 — switched to `raise ImportError(...)` to match the pattern in `ble.py` and `serial.py`.  Tests updated to expect `ImportError` (`ModuleNotFoundError` is a subclass, so callers catching either still work).

JPHutchins:
This has now been integrated and tested end-to-end in 2 downstream applications, including DFU testing (smpmgr and other).

davedesro (approved):

PR #110, opened 2026-05-29
docs: add Bumble transport documentation

ethan-mark-osu:
Resolves #109.

- Add docs/transport/bumble.md
- Add Bumble (BLE) entry to mkdocs.yaml nav

PR #111, opened 2026-05-30
Add Bumble documentation

### LINKED ISSUE #109: Add Bumble documentation
Update docs about transports. Explain how to get HCI FW as an extra.

YuuGR1337:
Fixes #109

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,12 @@ Or none (UDP transport only):
 smpclient
 ```
 
+With [Bumble](https://google.github.io/bumble/) transport for HCI forwarding:
+
+```
+smpclient[bumble]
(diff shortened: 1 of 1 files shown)


JPHutchins (changes requested):
Thanks! Might read better if it's added with the other transports. Those are also mentioned here, and that's what gets published.

https://github.com/intercreate/smpclient/blob/main/src/smpclient/__init__.py#L10-L18

JPHutchins:
Oh, and there's a new hci_firmware extra.

JPHutchins:
Closing because the implementation is too minimal, and #110 is more complete.

PR #114, opened 2026-06-11
Handle parsing protected TLV region

ChapterSevenSeeds:
Zephyr images built with TF-M get an MCUboot trailer generated with both protected and regular TLV entries. This PR modifies the MCUboot image parsing to properly recognize and parse these entries.


src/smpclient/mcuboot.py
```diff
@@ -325,18 +336,22 @@ def load_file(path: str) -> 'ImageInfo':
...
 
         f.seek(tlv_offset)  # move to the start of the TLV area
-        tlv_info = ImageTLVInfo.load_from(f)
 
-        tlvs: list[ImageTLVValue] = []
-        while f.tell() < tlv_offset + tlv_info.tlv_tot:
-            tlv_header = ImageTLV.load_from(f)
-            tlvs.append(ImageTLVValue(header=tlv_header, value=f.read(tlv_header.len)))
+        # The mcuboot design doc says that optional protected TLV entries come before regular TLV entries
+        protected_tlvs = []
```
JPHutchins: Presumably missing typing.

src/smpclient/mcuboot.py
```diff
@@ -301,6 +302,16 @@ def get_tlv(self, tlv: ImageTLVType) -> ImageTLVValue:
...
 
+    @staticmethod
+    def parse_tlvs(f: BytesIO, tlvs_end_offset: int) -> list[ImageTLVValue]:
+        """Parse TLVs from the current position in the file up to the given byte offset."""
+        tlvs: list[ImageTLVValue] = []
+        while f.tell() < tlvs_end_offset:
+            tlv_header = ImageTLV.load_from(f)
+            tlvs.append(ImageTLVValue(header=tlv_header, value=f.read(tlv_header.len)))
+
+        return tlvs
```
JPHutchins, line 327: I think that I agree with abstracting this to reduce repetition, but I don't love that it has the side effect of mutating the file read position. Consider inlining again (even though it's repetition) or documenting the side effect.

Maybe some cleverness can be done with an iterator or generator, probably not worth the effort.

JPHutchins (commented):
Thanks for this important update!

src/smpclient/mcuboot.py
ChapterSevenSeeds: I agree, I reworked it to take in a bytes object and just operate off of that.

JPHutchins (dismissed):
Thanks! LMK when you're ready to merge and I'll do so.

I think that linting must be disabled on this file, so instead of iterating, or asking you to rebase, I'm going to swing back around with a rework, confidence given by the integration test that was added.

ChapterSevenSeeds:
@JPHutchins Feel free to take it whenever convenient, though I am willing to continue trying to satiate the Microsoft AI beast if you'd like 😛

JPHutchins (approved):

PR #115, opened 2026-06-11
refactor(mcuboot): contain TLV info magics in the type system

JPHutchins:
## Summary

- `ImageMagic`, `ImageTLVInfoMagic`, and `ImageTLVProtInfoMagic` `Literal` type aliases tie each magic constant to its type (`Literal[...]` cannot reference a `Final` name per PEP 586, so the constants are typed `Final[<alias>]` to keep the value and type checked against each other).
- `ImageTLVInfo` is now generic over its magic, so protected and unprotected TLV info regions are distinct static types. The expected magic argument to `loads`/`load_from` binds the type parameter *and* supplies the runtime check, replacing the `protected: bool` flag. The loaders use a method-scoped TypeVar (`MagicT`) because class-scoped TypeVars only bind via `self`/`cls` — verified against both mypy and pyright.
- `ImageTLVInfo.load_tlvs_from(file)` reads and parses the TLV entries that follow the header, removing the `tlv_tot - IMAGE_TLV_INFO_STRUCT.size` wire-format arithmetic from `ImageInfo.load_file`.
- `ImageHeader` bad-magic validation moves from `__post_init__` (made unreachable by pydantic `Literal` validation) into `loads`, preserving `MCUBootImageError` on the load path.
- Tests: `assert_type` checks pin the static types at call sites, and new error-path tests bring `smpclient/mcuboot.py` to 100% line and branch coverage.

## Breaking changes

- `ImageTLVInfo.loads`/`load_from` take a required `magic` argument instead of `protected: bool = False`.
- `ImageTLVInfo.REGION_SIZE` is removed — use `IMAGE_TLV_INFO_STRUCT.size`.
- `ImageInfo.protected_tlvs` defaults to `None` instead of `[]`, pairing with `protected_tlv_info`.

## Test plan

- [x] `task format`, `task lint`, `task typecheck` (mypy), full `task test` — 296 passed
- [x] pyright clean on `mcuboot.py` and `test_mcuboot_tools.py` (loaders infer `ImageTLVInfo[Literal[0x6907]]`/`[Literal[0x6908]]`, no `Unknown`)
- [x] 100% line and branch coverage on `smpclient/mcuboot.py`

### DIFF
--- a/src/smpclient/mcuboot.py
+++ b/src/smpclient/mcuboot.py
@@ -3,20 +3,23 @@
 Specification: https://docs.mcuboot.com/design.html
 """
 
+from __future__ import annotations
+
 import argparse
 import pathlib
 import struct
 from enum import IntEnum, IntFlag, unique
 from functools import cached_property
 from io import BufferedReader, BytesIO
-from typing import Annotated, Any, Final, Union
+from typing import Annotated, Any, Final, Generic, Literal, TypeVar, Union
 
 from intelhex import hex2bin  # type: ignore
 from pydantic import Field, GetCoreSchemaHandler
 from pydantic.dataclasses import dataclass
 from pydantic_core import CoreSchema, core_schema
 
-IMAGE_MAGIC: Final = 0x96F3B83D
+ImageMagic = Literal[0x96F3B83D]
+IMAGE_MAGIC: Final[ImageMagic] = 0x96F3B83D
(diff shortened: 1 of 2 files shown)


JPHutchins:
FYI @ChapterSevenSeeds - stronger typing for mcuboot types, claude pass for "100% coverage".

PR #123, opened 2026-07-08
smpclient: upload: retry on timeouts

### LINKED ISSUE #56: Retransmit
Data can get lost - especially with transports like UDP. But even with a UART you could loose packets due to data corruptions. I noticed this while doing updates over UDP, where smpmgr(which uses smpclient) just fails on a a timeout every few packets.

I think, that we can handle this in a lower layer without modifying e.g. smpmgr, because the SMP protocol has sequence counters and thus can prevent repetitive execution of request handlers - which could have unintended side-effects.

What do you think?

M1cha:
This is important for unrealiable transports like UDP over OpenThread.

Fixes: https://github.com/intercreate/smpclient/issues/56


M1cha:
I don't understand the CI error. Why would it think that `response` doesn't have those fields anymore?

src/smpclient/__init__.py
```diff
@@ -324,20 +326,33 @@ async def upload(
...
-            )
+                    timeout_s=subsequent_timeout_s,
+                )
+                current_attempt = 0
+            except TimeoutError:
+                current_attempt += 1
+                continue
+            except SMPBadSequence:
+                current_attempt += 1
+                continue
```
JPHutchins, line 354: The type error happens here where it loops back around before checking for errors.

JPHutchins:
@M1cha I appreciate you picking this one up again! I'm wondering if you'd be interested in generalizing it? The reason is that it's not only UDP or image uploads that suffer from this defect, so it would be nice to fix this all at once.

I've posted an implementation that I never finished here, if you'd like to purse that approach: https://github.com/intercreate/smpclient/issues/56#issuecomment-4917117914

It uses the sequence value to synchronize and drop stale messages. But I think it needs unit testing, integration testing, and field testing. I tested it by verifying that I could unplug/replug UART lines during an upload and that it would recover without starting over.

Issue #103, opened 2026-04-01
Add HCI option to BLE Transport


JPHutchins:
Users may like to select their HCI rather than using Bleak default.

_Originally posted by @hseem in https://github.com/intercreate/smpclient/issues/30#issuecomment-4170726778_

Issue #90, opened 2025-12-07
Bleak transport options


JPHutchins:
https://github.com/intercreate/smpmgr/discussions/82

JPHutchins:
> [!WARNING]
> **LLM Disclosure**
>
> This comment was authored by claude-opus-4-8 on behalf of @JPHutchins, coordinating this work with the cross-repo migration.

Staging note for the transport-subcommands epic (intercreate/smpmgr#103): this lands on `smpclient`'s `screaming-goblin` branch, which will **pin `smp`'s `screaming-goblin`** (git dependency) so smp → smpclient → smpmgr migrate in lockstep before `smp` cuts its breaking release; the pin then moves to the released version.

The `smp` side is progressing on its own `screaming-goblin` (Frame[T]/Data + generic request/response now in `smp`); see the epic for the rollup.

PR #13, opened 2024-03-31
Example/ble dfu

JPHutchins:
@tomaszduda23 I have received the nRF52DK (accidentally got nRF52832, nRF52840 is on the way), the Adafruit Feather nRF52840 and the Pro Micro nRF52840.

This PR does a bit of housekeeping around `mypy`.  Ignore that 😭.

But it also adds and examples folder and some DUT routines for the nRF52DK and Feather!  I hope that you'll be able to test these with your hardware so that we can get to the bottom of any BLE OTA DFU issues.

Here's an example output of the feather upgrade routine.  For the sake of simplicity I've tested from Windows, but I'd hope that the BLE from Linux would mostly work that same.  I included your workaround for BLE MTU and have mentioned it on another open ticket over at Bleak.

```
python -m examples.ble.upgrade adafruit_feather_nrf52840                                                                                                           
Using DUT folder: C:\Users\jp\repos\smpclient\examples\duts\adafruit_feather_nrf52840
Flashing the merged.hex...
[ #################### ]   1.140s | Erase file - Done erasing
[ #################### ]   1.581s | Program file - Done programming
[ #################### ]   1.629s | Verify file - Done verifying
Applying system reset.
Run.
A SMP DUT hash: SHA256=a3837de56f67f91d8e07cf87f3dba194ca9284c46cd8946c51bc64f58a26a4f9
B SMP DUT hash: SHA256=f9faba4f03ae60601f5df33d106f3714a7c074f57ad8cb700c011591e97d57f2
Searching for A SMP DUT...OK
Connecting to A SMP DUT...OK
Sending request...OK
Received response: header=Header(op=<OP.READ_RSP: 1>, version=<Version.V0: 0>, flags=<Flag: 0>, length=134, group_id=<GroupId.IMAGE_MANAGEMENT: 1>, sequence=0, command_id=<ImageManagement.STATE: 0>) sequence=0 images=[ImageState(slot=0, version='0.0.0', image=None, hash=b'\xa3\x83}\xe5og\xf9\x1d\x8e\x07\xcf\x87\xf3\xdb\xa1\x94\xca\x92\x84\xc4l\xd8\x94lQ\xbcd\xf5\x8a&\xa4\xf9', bootable=True, pending=False, confirmed=True, active=True, permanent=False)] splitStatus=0

Uploaded 223,630 / 223,630 Bytes
Sending request...OK
[...]

PR #144, opened 2026-09-23
breaking: the effect-stack lifecycle (#58): transports own their link, options, and sizing

JPHutchins:
> [!WARNING]
> **LLM Disclosure**
>
> This PR was authored by `claude-opus-5-5[1m]` on behalf of @JPHutchins, implementing the #58 effect-stack design we iterated on (https://github.com/intercreate/smpclient/pull/58#issuecomment-5786045349) as a single PR, per @JPHutchins.

> [!NOTE]
> Targets **`screaming-goblin`**, the breaking branch (#143, epic intercreate/smpmgr#103). Closes #90. Closes #103.

**The transport owns its address, options, link and sizing; `SMPClient` owns nothing but its sequence space.** Each transport is obtained already connected and sized from its own `async with` bracket, so there is exactly one bracket and `SMPClient(transport)` is a plain object.

```python
async with SMPSerialTransport.connected("/dev/ttyACM0") as transport:  # opens, then reads the params (Auto)
    client = SMPClient(transport)
    await client.upload(image)

async with SMPBLETransport.borrow(bleak_client) as transport:  # @GHF's standing link: never opened or closed here
    client = SMPClient(transport)
```

- `SMPTransport` is now just the live interface: `send`, `receive`, `send_and_receive`, `mtu`, `max_unencoded_size`. Nothing on it opens or closes a link, so a client can't drop a link it doesn't own. That holds by construction and is checked by mypy and pyright.
- Constructors wrap an **already-open** resource. `Transport.connected(...)` opens one; `Transport.borrow(...)` wraps the caller's.
- MCUmgr params negotiation is each transport's **opt-in sizing policy**: a per-transport sum type sharing one `Auto()`. Only `Auto()` (and GATT `Unfragmented()`) sends the params read, inside the bracket before it yields. An error response falls back with a warning; a timeout raises.

<details>
<summary>Breaking changes</summary>

| before | after |
|---|---|
| `async with SMPClient(transport, address) as c` | `async with Transport.connected(address, ...) as t:` then `c = SMPClient(t)` |
| `SMPClient.connect()` / `.disconnect()` / `.address` | removed |
| `SMPTransport.connect` / `disconnect` / `initialize` | removed from the Protocol |
| `SMPUDPTransport(mtu)` + `connect(addr, t, port=1337)` | `SMPUDPTransport.connected(address, port=1337, *, mtu=, sizing=)` |
| `SMPBLETransport(winrt=)` | `SMPBLETransport.connected(address, *, winrt=, bluez=, sizing=)` / `.borrow(client)` |
| `SMPBumbleTransport(hci=, ...)` + `connect` / `use_connection` / `borrowed_connection` | `SMPBumbleTransport.connected(address, *, hci=, ...)` / `.borrow(connection)` |
| `SMPBumbleTransport.bonded_devices` / `clear_bond` / `clear_bonds` | module functions `bonded_devices(keystore=, host_address=)` etc. |
| `SMPSerialTransport(fragmentation_strategy=, <pyserial kwargs>)` | `SMPSerialTransport.connected(port, *, sizing=, line_length=, options=SerialOptions(...))` |
| `FragmentationStrategy` | `SerialSizing` |
| `BufferSize(buf_size, line_length)` / `BufferParams(line_length, line_buffers)` | `BufferSize(buf_size)` (shared by every transport) / `BufferParams(line_buffers)`; `line_length` is a serial transport option |
| `SMPSerialRawTransport(mtu=384)` | `sizing=Auto()` (384 until negotiated) or `BufferSize(n)` |
| the deprecated 7.1.0 serial params (`max_smp_encoded_frame_size`, ...) | removed |
| a params timeout under default sizing | now raises (was logged and ignored) |

</details>

<details>
<summary>Decisions made while implementing, for review</summary>

- **One shared `BufferSize(buf_size)`.** Raw serial lives in the same `serial` package and needs a plain pinned size, so two different `BufferSize` types there would be a trap. Serial's `line_length` therefore moved onto the transport, which also lets `Auto` set it (it couldn't before).
- **`SerialPort` Protocol** (`port`, `out_waiting`, `write`, `read_all`) types the serial `conn`. pyserial's `serial_for_url` ports, including the emulator `socket://` chardev the integration suite uses, are `SerialBase`s, not `serial.Serial`. The old harness hid that behind `object.__setattr__`; now it just calls `SMPSerialTransport.borrow(chardev)`.
- **`SerialOptions`** NamedTuple replaces the 11 pyserial kwargs that were repeated across three signatures.
- **`Unfragmented()` also reads the params.** It is `Auto` capped at the ATT write size, for a server without `CONFIG_MCUMGR_TRANSPORT_BT_REASSEMBLY`, and it needs `buf_size` to stay under a netbuf smaller than the MTU.
- **bleak `borrow()` polls `is_connected`** for disconnects: bleak accepts a disconnect callback only at construction, and the owner's callback must not be replaced.
- **UDP has no `borrow()`**, since its client class is private. The integration suite's skip for a non-default UDP port is gone, because the port is a real parameter now.
- The request/response exchange moved from `SMPClient.request` to a private `smpclient._request`, so transports can read the params themselves. `success`/`error`/`error_v1`/`error_v2` are re-exported from `smpclient` unchanged.
- The integration harness waits on a pinned `BufferSize(384)` transport until a booting server answers, then re-borrows the same link with `Auto()`. Negotiation happens once, against a server that is up.

</details>

<details>
<summary>Verification</summary>

- `camas check`: ruff, pydoclint, mypy, pyright, and tests all green (441 passed, 14 skipped; coverage 95.4%, floor 91%).
- `camas matrix`: green on 3.10, 3.11, 3.12, 3.13, 3.14.
- `camas test_integration` on the final tip: **229 passed**, the same 229 as #137's baseline. That includes serial recovery over all four framings and the UDP fixtures, and every one of the 101 skips is a by-design transport or reliability filter.
- The two negative properties are enforced by the type checkers: you can't construct an unconnected transport (the constructor needs the open resource), and a client can't close a link (`SMPTransport` has no `__aexit__`).
- Porting the tests caught four `src` issues, each fixed and pinned by a test that fails without the fix:
  1. `Auto` with `line_length < 22` couldn't send the params request.
  2. bleak `connect_timeout_s` had stopped covering `start_notify`.
  3. bumble `send()` after exit wrote to a torn-down link.
  4. A cancel mid-`connected()` skipped bumble's teardown.
- Commits are split by layer for review. Only the tip is gated: the Protocol break makes the intermediate commits red on their own.

</details>


src/smpclient/transport/serial/__init__.py
```diff
@@ -3,11 +3,14 @@
 In addition to UART, these transports can be used with USB CDC ACM and CAN.
 """
 
-from smpclient.transport.serial.encoded import Auto as Auto
+from smpclient.transport import Auto as Auto
```
JPHutchins, line 6: Why re-export these?

JPHutchins: The existing re-exports were needed becuase of the encoded/unencoded split

JPHutchins:
> [!WARNING]
> **LLM Disclosure**
>
> This review was authored by `claude-opus-5-5[1m]` on behalf of @JPHutchins. JP found the PR more invasive than expected and asked for three passes, in order: an advisor review, a full `/code-review` at `xhigh`, and an advisor review of that output. This comment posts the outcome.
>
> **Read it as a self-review.** The implementer and the `/code-review` agent are the same model (`claude-opus-5-5`). The advisor passes ran on `claude-fable-5-1`.

## Summary

- **19 findings:** 15 from `/code-review`, plus 4 that only the advisor raised. The happy path holds: CI is 28/28 and integration is 229 passed. The bugs are all on teardown, cancel and misuse paths.
- **9 are real bugs to fix before merge.** The worst four:
  - bumble `borrow()` teardown turns off *everyone's* notifications.
  - bleak's exit path can hang, the same class of failure as intercreate/smpmgr#97.
  - an import shadows `smpclient._request`, which breaks `mock.patch` on 3.10.
  - `pair(force=True)` after the bracket exits deletes a valid bond.
- **4 decisions are yours:**
  - the cost of the timeout policy
  - the sequence number the params read uses
  - the `FragmentationStrategy` naming
  - whether to split the PR
- **Nothing is changed yet.** The naming change will churn the diff anyway, so fixes wait for your answers.

<details>
<summary><b>Bugs to fix (9), each with a proposed fix</b></summary>

| # | Where | Bug | Proposed fix |
|---|---|---|---|
| 1 | `bumble/__init__.py` `_teardown_borrowed` | `smp_characteristic.unsubscribe()` with no subscriber drops **every** subscriber on the handle and writes 0 to the CCCD. Verified in bumble: `Client.unsubscribe(subscriber=None)` pops them all. A caller's own subscription, or an outer `borrow()`, goes silent, and the proxy entry leaks. This predates the PR but is now exposed, because `borrow()` makes a shared link the intended use. | Keep the `partial` and pass it: `unsubscribe(subscriber)`. |
| 2 | `ble.py` `_subscribed` finally | `await client.stop_notify()` runs on every exit with no timeout and no disconnect guard. If the peer is gone but BlueZ still reports connected, exit hangs forever. On the owned path it's redundant, since `disconnect()` follows. | Owned path: skip `stop_notify`. `borrow()`: guard it with `_await_or_disconnect` plus a timeout. |
| 3 | `ble.py` `connected()` finally | A bare `await client.disconnect()` can raise, and the establishment error that mattered gets replaced. | Make it best-effort, as the old `_best_effort_disconnect` was. |
| 4 | `smpclient/__init__.py:61` | `from smpclient._request import request as _request` rebinds the submodule attribute to the function. `import smpclient._request as m` then gives a function, and on 3.10 `mock.patch("smpclient._request.…")` patches the function instead of the module. | Import the module: `from smpclient import _request`. |
| 5 | `bumble/__init__.py` `pair()` | The removed state guard was doing real work. After exit, `pair(force=True)` deletes the stored bond (verified: `pairing.pair` deletes before pairing), then fails on the dead link. | The same disconnected guard `send()` has. |
| 6 | `serial/common.py` `open_serial` | `reset_input_buffer()` is inside the retry `try`. If it raises after `open()`, the retries hit "already open" and the file descriptor leaks. | Move it after the loop, inside the `try/finally`. |
| 7 | `serial/encoded.py` `BufferParams` | The first positional field changed from `line_length` to `line_buffers`, and both are `int`. Old code `BufferParams(256)` now silently means 256 line buffers. | Keyword-only fields, e.g. `@dataclass(frozen=True, kw_only=True)`. |
| 8 | `smpclient/__init__.py` `SMPClient.__init__` | `timeout_s` sits where `address` was, so un-migrated `SMPClient(t, "COM3")` constructs fine and fails later inside `timeout()`. | Make `timeout_s` and `sequence` keyword-only. |
| 9 | `bumble/__init__.py` `connected()` except | `(Exception, asyncio.CancelledError)` goes to `logger.exception`, so an ordinary cancel or outer timeout logs an ERROR traceback. | A separate `CancelledError` arm, logged at debug, then re-raised. |

**Smaller items:**
- The `encoded.py` module docstring says "`Auto` assumes two 128-byte lines". That's a specific claim, and it's wrong for a non-default `line_length`. Make it vague.
- Tests compare sizing NamedTuples by value, as in `t._sizing == PINNED`, which also passes for `BufferSize(4)`. Use `isinstance` plus the field.
- In `_request.request(transport, request, …)`, the parameter shadows the function's own name.
- bleak `_watch_link` logs a WARNING "Disconnected" when the owner closes a borrowed link on purpose. That level is wrong for that path.
- `Unfragmented` lives in shared `smpclient.transport`, though only the GATT transports accept it.
- `connect_timeout_s` bounds each step separately: 2× for serial, 3× for BLE. Serial's docstring reads as a single bound.
- The harness's `answering()` re-borrows a serial port it hasn't flushed. That's safe only because the echo is fully consumed first; it needs a comment, and see B below.

</details>

<details>
<summary><b>Your decisions (4)</b></summary>

**A. The cost of the timeout policy** (`/code-review` #1). This is implemented as you specified it: "fall back on error, raise on timeout".
- **What it costs, per the review:** a freshly reset device, whose USB CDC ACM is up before the SMP server is, now fails on entry. So does a UDP link that drops the one params datagram. Before, those carried on at the fallback size. The integration harness needed a probe-then-negotiate pattern for exactly this reason, and every caller talking to a just-reset device will need the same.
- **Options:**
  1. Keep it.
  2. Raise the default `connect_timeout_s`.
  3. Add an explicit "negotiate, else fall back" variant.
  4. Retry the read inside the bracket.

**B. The sequence number of the params read** (review #2 and #12, advisor). The read hardcodes sequence 0, outside the client's injectable sequence space, which is the principle smp#71 set. `wrapping_sequence()` also starts at 0, so the params read and the client's first request are consecutive seq-0 frames.
- **Consequence:** on a serial port that hasn't been flushed, a stale reply shows up as `SMPValidationException`, not the `SMPBadSequence` that `_poll_until_answering` tolerates. That's a plausible latent flake in the harness.
- **Proposal:** `connected()`/`borrow()` take a `sequence: Iterator[u8]` (default: a fresh `wrapping_sequence()`), and the negotiation consumes its first value.

**C. Naming** (your call during review). Rename the concept to `FragmentationStrategy`, with one union per transport and the kwarg back to `fragmentation_strategy`. Proposed names:

| Transport | Union | Options |
|---|---|---|
| console serial | `SerialFragmentationStrategy` | `Auto \| BufferSize \| BufferParams` |
| raw serial | `RawSerialFragmentationStrategy` | `Auto \| BufferSize` |
| BLE (bleak, bumble) | `GATTFragmentationStrategy` | `Auto \| Unfragmented \| BufferSize` |
| UDP | `UDPFragmentationStrategy` | `Auto \| BufferSize` |

The prefixes are there because the console and raw transports share the `serial` package, so a bare `FragmentationStrategy` per module would clash.

**D. Split the PR?** Most of the churn is forced by the design:
- the `_request.py` extraction (~220 lines)
- the bumble closure restructure (~630 lines of churn, mostly moved rather than changed)
- the ported tests (+2142, much of it new coverage)

What could come out, if you want a smaller PR:
- the deprecated 7.1.0 serial params removal (~200 lines, unrelated to #58)
- the bumble bond functions becoming module-level
- `SerialOptions`

</details>

<details>
<summary><b>Public surface you haven't approved yet</b></summary>

| Surface | Why it exists | Question |
|---|---|---|
| `SerialOptions` | Groups the 11 pyserial settings that were repeated across three signatures. | Keep, or go back to flat kwargs? |
| `SerialPort` Protocol | pyserial's `serial_for_url` ports, including the emulator's `socket://` chardev, are a `SerialBase`, not a `serial.Serial`. | Keep public? |
| `open_serial` in `serial.common` | The retry-until-open bracket. The integration harness imports it from `common`. | Re-export from `smpclient.transport.serial`, or keep it internal? |
| bumble `bonded_devices` / `clear_bond` / `clear_bonds` as module functions | They need a keystore and a host address, never a connection. | OK? |
| `Connected` / `ConnectedBorrowed` as `SMPBumbleTransport(...)` arguments | The constructor wraps a live link. | Callers should use the brackets. Make them private? |
| `SMPBLETransport.__init__(client, characteristic, max_write, disconnected, …)` | Same reason. | Same. |

</details>

<details>
<summary><b>Pushed back on (1)</b></summary>

**`/code-review` #13:** `_negotiate` is copied into five transports. That's true, but it follows from the round-2 design, where the policy is one exhaustive `match` inside each transport. A shared `negotiate_buf_size() -> int | None` would move the policy back out of the transport. I recommend keeping the duplication.

</details>

**Next:** once A–D are answered, one fix commit for bugs 1–9 and the smaller items, then the naming rename, then re-gate and re-run integration.

src/smpclient/transport/serial/__init__.py
JPHutchins: > [!WARNING]
> **LLM Disclosure**
>
> This reply was authored by `claude-opus-5-5[1m]` on behalf of @JPHutchins. It answers this thread for the force-pushed revision of the PR.

These re-exports are gone in the new revision. [`smpclient/transport/serial/__init__.py`](https://github.com/intercreate/smpclient/blob/d35531e289ca9cbae80188bfd2ad4f7da7028442/src/smpclient/transport/serial/__init__.py) no longer re-exports any shared `smpclient.transport` type. `Auto` is imported from `smpclient.transport`, as are the shared `BufferSize` and `Unfragmented`.

What it does export are serial's own types, defined in its `encoded`, `unencoded`, and `common` submodules. That is the same split-driven reason as before: serial's own `BufferSize` (unchanged from `main`), plus the new `SerialOptions` and `SerialPort` from `common`.

JPHutchins:
> [!WARNING]
> **LLM Disclosure**
>
> This comment was authored by `claude-opus-5-5[1m]` on behalf of @JPHutchins. It marks the force-push of the less invasive revision she asked for.

**Force-pushed the new revision:** `a76b962` → `d35531e`. It is a fresh branch of 11 commits, each passing `camas check` on its own, and the description is rewritten for it. The [earlier review](https://github.com/intercreate/smpclient/pull/144#issuecomment-5787501511) applies to `a76b962`.

| | src | tests |
|---|---|---|
| `a76b962` | +1424 / −1209 | +2142 / −1152 |
| `d35531e` | +1029 / −601 | +880 / −430 |

At the tip:
- `camas check`: 401 passed, 14 skipped; coverage 94.1%.
- `camas matrix`: green on 3.10–3.14.
- `camas test_integration`: 229 passed, 101 skipped, the same as the baseline.

src/smpclient/transport/bumble/__init__.py
```diff
@@ -160,7 +173,17 @@ def __init__(
                 `pair_on_connect` and `pair()`.
             settle_s: Wait between successful pair and proceeding (or
                 disconnecting) so the peer can finalize bonding.
+            fragmentation_strategy: How to size SMP messages: `Auto`, `Unfragmented`, or
+                `BufferSize`.
+            connect_timeout_s: Bounds scanning for a name, and reading the server's
+                MCUmgr parameters.
+            sequence: The SMP sequence space the MCUmgr parameters read draws from;
+                defaults to `wrapping_sequence()`.
```
JPHutchins: Yu say it defaults to wrapping_sequence, yet you default it to None in the sig

JPHutchins: The correct approach is for it to be structural (type level) not "prose".

src/smpclient/transport/bumble/__init__.py
```diff
@@ -160,7 +173,17 @@ def __init__(
...
         """
+        self._address: Final = address
+        self._fragmentation_strategy = fragmentation_strategy
+        self._connect_timeout_s = connect_timeout_s
+        self._sequence = _request.wrapping_sequence() if sequence is None else sequence
```
JPHutchins: Are these final?

src/smpclient/transport/bumble/__init__.py
```diff
@@ -555,6 +587,35 @@ async def _teardown_borrowed(self) -> None:
             logger.warning(f"remove_listener(EVENT_DISCONNECTION) failed: {e}")
 
 
+async def bonded_devices(
+    *, keystore: KeystoreStrategy = Tempfile(), host_address: Address = DEFAULT_HOST_ADDRESS
```
JPHutchins: why kwargs only?

src/smpclient/transport/serial/common.py
```diff
@@ -14,105 +17,197 @@
...
-            xonxoff=xonxoff,
-            rtscts=rtscts,
-            write_timeout=write_timeout,
-            dsrdtr=dsrdtr,
-            inter_byte_timeout=inter_byte_timeout,
-            exclusive=exclusive,
-        )
+        self._port: Final = port
+        self._connect_timeout_s = connect_timeout_s
+        self._sequence = _request.wrapping_sequence() if sequence is None else sequence
```
JPHutchins: Are these final?

src/smpclient/transport/serial/encoded.py
```diff
@@ -145,8 +140,13 @@ class BufferParams(NamedTuple):
...
 
-FragmentationStrategy: TypeAlias = Auto | BufferSize | BufferParams
-"""How `SMPSerialTransport` sizes SMP messages: `Auto`, `BufferSize`, or `BufferParams`."""
+SerialFragmentationStrategy: TypeAlias = Auto | BufferSize | BufferParams
+"""How `SMPSerialTransport` sizes SMP messages: `Auto`, `BufferSize`, or `BufferParams`.
+
+With `Auto`, connecting reads the server's `buf_size` (the decoded reassembly buffer) and
+the transport sends messages up to `buf_size - 4`, filling that buffer; until the parameters
+are read, or if the server doesn't provide them, it assumes a conservative line budget.
+"""
```
JPHutchins: llm doc slop - restates the code and other doc strings

src/smpclient/transport/serial/encoded.py
```diff
@@ -190,19 +190,12 @@ class BufferState(IntEnum):
     @overload
     def __init__(
         self,
-        fragmentation_strategy: FragmentationStrategy = ...,
+        port: str,
```
JPHutchins: Port wouldn't be required if we achieve via SMPSerialTransport.borrowed()... right?

src/smpclient/transport/serial/encoded.py
```diff
@@ -236,79 +222,52 @@ def __init__(
...
-        rtscts: bool = False,
-        write_timeout: float | None = None,
-        dsrdtr: bool = False,
-        inter_byte_timeout: float | None = None,
-        exclusive: bool | None = None,
+        connect_timeout_s: float = 2.5,
+        sequence: Iterator[u8] | None = None,
+        options: SerialOptions = SerialOptions(),
     ) -> None:
         """Initialize the serial transport.
```
JPHutchins: The overlaods were just for backwards compat. This is a breaking change, good time to drop this bullshit.

src/smpclient/transport/__init__.py
```diff
@@ -62,14 +83,6 @@ async def send_and_receive(self, data: bytes) -> bytes:  # pragma: no cover
         """
         ...
 
-    def initialize(self, smp_server_transport_buffer_size: int) -> None:  # pragma: no cover
-        """Initialize the `SMPTransport` with the server transport buffer size.
-
-        Args:
-            smp_server_transport_buffer_size: The SMP server transport buffer size, in 8-bit bytes.
-        """
-        self._smp_server_transport_buffer_size = smp_server_transport_buffer_size
```
JPHutchins: Isn't negotiate required now?

JPHutchins: Errr, I guess not, since client doesn't call it.

src/smpclient/transport/__init__.py
```diff
@@ -88,5 +101,85 @@ def max_unencoded_size(self) -> int:  # pragma: no cover
...
+    """The server's advertised `buf_size`, once a fragmentation strategy that asks has read it."""
 
-        return self._smp_server_transport_buffer_size or self.mtu
+    async def connect(self) -> None:  # pragma: no cover
+        """Open the link, then `negotiate()`."""
+        ...
+
+    async def disconnect(self) -> None:  # pragma: no cover
+        """Close the link."""
+        ...
```
JPHutchins: Add note to not use them; dangerous loss of side effect control

src/smpclient/transport/ble.py
```diff
@@ -81,63 +91,159 @@ class SMPBLETransportNotSMPServer(SMPBLETransportException):
...
+        fragmentation_strategy: GATTFragmentationStrategy = Auto(),
+        connect_timeout_s: float = 2.5,
+        sequence: Iterator[u8] | None = None,
+    ) -> None:
+        """Initialize the BLE transport; `connect()` scans for and connects to `address`.
+
+        Args:
+            address: The device's MAC address, macOS UUID, or advertised name.
+            winrt: WinRT backend arguments, e.g. `use_cached_services`.
+            bluez: BlueZ backend arguments, e.g. the `adapter` to scan and connect with.
```
JPHutchins: You're telling me that we accpet both? On macos too? At the same time. FUCKING GARBAGE. This is a sum type.

src/smpclient/transport/ble.py
```diff
@@ -81,63 +91,159 @@ class SMPBLETransportNotSMPServer(SMPBLETransportException):
...
+                `BufferSize`.
+            connect_timeout_s: Bounds scanning and connecting, and reading the server's
+                MCUmgr parameters.
+            sequence: The SMP sequence space the MCUmgr parameters read draws from;
+                defaults to `wrapping_sequence()`.
+        """
+        self._address: Final = address
+        self._fragmentation_strategy = fragmentation_strategy
+        self._connect_timeout_s = connect_timeout_s
+        self._sequence = _request.wrapping_sequence() if sequence is None else sequence
```
JPHutchins: Final?

src/smpclient/__init__.py
```diff
@@ -141,49 +77,8 @@ def success(
...
-            *(
-                f"\tCould not be parsed as {response.__name__}: {error}"
-                for response, error in errors
-            ),
-        )
-    )
-    return summary, details
-
-
 class SMPClient:
```
JPHutchins: Should be generic over the transport, so that users can access the transport in a type safe way.

JPHutchins (commented):
Going the right direction. Overall poor code quality, needs some review and revision. Better typing, less mutation, etc etc.

src/smpclient/transport/bumble/__init__.py
JPHutchins: > [!WARNING]
> **LLM Disclosure**
>
> This reply was authored by `claude-opus-5-5[1m]` on behalf of @JPHutchins, addressing her review of `5d58791`.

Fixed in [`e4e8228`](https://github.com/intercreate/smpclient/pull/144/commits/e4e8228d685585ce8f395957ad70b852d271add5). The default is now in the signature, `sequence: Callable[[], Iterator[u8]] = wrapping_sequence`, on every transport and on `SMPClient`, and the "defaults to" prose is gone. It's a factory rather than an iterator because a default iterator is evaluated once at definition time and would be shared by every instance.

JPHutchins: > [!WARNING]
> **LLM Disclosure**
>
> This reply was authored by `claude-opus-5-5[1m]` on behalf of @JPHutchins, addressing her review of `5d58791`.

Agreed; it's structural now ([`e4e8228`](https://github.com/intercreate/smpclient/pull/144/commits/e4e8228d685585ce8f395957ad70b852d271add5)): `sequence: Callable[[], Iterator[u8]] = wrapping_sequence`.

src/smpclient/transport/bumble/__init__.py
JPHutchins: > [!WARNING]
> **LLM Disclosure**
>
> This reply was authored by `claude-opus-5-5[1m]` on behalf of @JPHutchins, addressing her review of `5d58791`.

They are now ([`e4e8228`](https://github.com/intercreate/smpclient/pull/144/commits/e4e8228d685585ce8f395957ad70b852d271add5)). As Protocol members they couldn't be: a `Final` assignment in the implementer conflicts with the writable Protocol attribute. `_ConnectableTransport` is now a concrete base, generic over the strategy union, and its `__init__` sets `_fragmentation_strategy`, `_connect_timeout_s`, and `_sequence` as `Final`.

src/smpclient/transport/bumble/__init__.py
JPHutchins: > [!WARNING]
> **LLM Disclosure**
>
> This reply was authored by `claude-opus-5-5[1m]` on behalf of @JPHutchins, addressing her review of `5d58791`.

No reason; dropped in [`7c91b0b`](https://github.com/intercreate/smpclient/pull/144/commits/7c91b0b922e0dc2906c328940ee0a2a07ad2faa9). `bonded_devices`, `clear_bond`, and `clear_bonds` take their arguments positionally too.

src/smpclient/transport/serial/common.py
JPHutchins: > [!WARNING]
> **LLM Disclosure**
>
> This reply was authored by `claude-opus-5-5[1m]` on behalf of @JPHutchins, addressing her review of `5d58791`.

Yes, now ([`e4e8228`](https://github.com/intercreate/smpclient/pull/144/commits/e4e8228d685585ce8f395957ad70b852d271add5)), through the concrete base's `__init__`. `_port` is gone entirely ([`f6678a1`](https://github.com/intercreate/smpclient/pull/144/commits/f6678a1d5cc6a92474b0d547e31ee711b8651bb5)): the port is an argument to `connect(port)`.

src/smpclient/transport/serial/encoded.py
JPHutchins: > [!WARNING]
> **LLM Disclosure**
>
> This reply was authored by `claude-opus-5-5[1m]` on behalf of @JPHutchins, addressing her review of `5d58791`.

Cut in [`7c91b0b`](https://github.com/intercreate/smpclient/pull/144/commits/7c91b0b922e0dc2906c328940ee0a2a07ad2faa9): `SerialFragmentationStrategy` is one line. So are the raw and UDP aliases. The constructors' `fragmentation_strategy` Args no longer repeat the union.

src/smpclient/transport/serial/encoded.py
JPHutchins: > [!WARNING]
> **LLM Disclosure**
>
> This reply was authored by `claude-opus-5-5[1m]` on behalf of @JPHutchins, addressing her review of `5d58791`.

Right. The target moved to the primitive that uses it ([`f6678a1`](https://github.com/intercreate/smpclient/pull/144/commits/f6678a1d5cc6a92474b0d547e31ee711b8651bb5)): `SMPSerialTransport(BufferSize(1024)).connected("/dev/ttyACM0")` opens, `SMPSerialTransport().borrowed(port)` adopts, and the constructor holds only config. The same holds for BLE and bumble (`connected(address)`) and UDP (`connected(address, port=1337)`).

src/smpclient/transport/serial/encoded.py
JPHutchins: > [!WARNING]
> **LLM Disclosure**
>
> This reply was authored by `claude-opus-5-5[1m]` on behalf of @JPHutchins, addressing her review of `5d58791`.

Dropped in [`a1c2b16`](https://github.com/intercreate/smpclient/pull/144/commits/a1c2b1666428f2ff2c5f13e36815f0ea1601ba60): the three overloads, `_LegacyParams`, `_resolve_fragmentation_strategy`, the deprecation constants, and their tests. The only signature is `SMPSerialTransport(fragmentation_strategy=Auto(), *, connect_timeout_s, sequence, options)`.

src/smpclient/transport/__init__.py
JPHutchins: > [!WARNING]
> **LLM Disclosure**
>
> This reply was authored by `claude-opus-5-5[1m]` on behalf of @JPHutchins, addressing her review of `5d58791`.

Right, it isn't: `negotiate()` is abstract on `_ConnectableTransport` and absent from `SMPTransport`, so `SMPClient` can't call it. `connect()` and `borrow()` run it, and it stays public for re-negotiating a live link, for example after a boot race or a firmware update.

src/smpclient/transport/__init__.py
JPHutchins: > [!WARNING]
> **LLM Disclosure**
>
> This reply was authored by `claude-opus-5-5[1m]` on behalf of @JPHutchins, addressing her review of `5d58791`.

Added. The `_ConnectableTransport` docstring now reads: "Open a link with the `connected()` or `borrowed()` bracket: the bare `connect()`, `borrow()`, and `disconnect()` primitives give up the bracket's guarantee that the link is released on error and on cancellation." Each primitive's docstring ends in "prefer `connected()`" or "prefer `borrowed()`" ([`e4e8228`](https://github.com/intercreate/smpclient/pull/144/commits/e4e8228d685585ce8f395957ad70b852d271add5), [`f6678a1`](https://github.com/intercreate/smpclient/pull/144/commits/f6678a1d5cc6a92474b0d547e31ee711b8651bb5)). The wording says *prefer* rather than *don't use*; say if you want it stronger.

src/smpclient/transport/ble.py
JPHutchins: > [!WARNING]
> **LLM Disclosure**
>
> This reply was authored by `claude-opus-5-5[1m]` on behalf of @JPHutchins, addressing her review of `5d58791`.

Agreed; it's one sum type now ([`eab0363`](https://github.com/intercreate/smpclient/pull/144/commits/eab036305673b890a1bf97df97565bd688b3f23e)): `backend: BleakBackend = PlatformDefault()`, where `BleakBackend = PlatformDefault | BlueZ | WinRT`, and exactly one backend's options reach bleak. The variants wrap bleak's own TypedDicts, `BlueZ(BlueZClientArgs(adapter="hci1"))`, so bleak's schema stays the single source of truth. The cost is one extra layer compared with a flat `BlueZ(adapter="hci1")`; say if you'd rather have the flat form.

src/smpclient/transport/ble.py
JPHutchins: > [!WARNING]
> **LLM Disclosure**
>
> This reply was authored by `claude-opus-5-5[1m]` on behalf of @JPHutchins, addressing her review of `5d58791`.

Yes ([`e4e8228`](https://github.com/intercreate/smpclient/pull/144/commits/e4e8228d685585ce8f395957ad70b852d271add5)): the config through the base's `__init__`, and `_buffer`/`_notify_condition`/`_disconnected_event`/`_backend` in the constructor.

src/smpclient/__init__.py
JPHutchins: > [!WARNING]
> **LLM Disclosure**
>
> This reply was authored by `claude-opus-5-5[1m]` on behalf of @JPHutchins, addressing her review of `5d58791`.

Done in [`7bad118`](https://github.com/intercreate/smpclient/pull/144/commits/7bad118a3782c2e914bb0ea5508b99856f820bf9): `SMPClient(Generic[TTransport])` with `TTransport = TypeVar("TTransport", bound=SMPTransport)`, and a `transport` property typed `TTransport`. `SMPClient(SMPSerialTransport()).transport.read_serial()` type-checks with no cast; `ICUploadClient` is generic too. `tests/test_generics_typing.py` asserts it under mypy and pyright.

JPHutchins:
> [!WARNING]
> **LLM Disclosure**
>
> This comment was authored by `claude-opus-5-5[1m]` on behalf of @JPHutchins. It summarizes the fix-ups for her review of `5d58791` ("better typing, less mutation"), done against her coding guidelines.

**Fast-forwarded `5d58791` → `7c91b0b`: six fix-up commits, no force-push.** Each commit passes `camas check`. At the tip, the matrix passes on 3.10–3.14 and integration matches the baseline at 229 passed, 101 skipped. Every thread has a reply naming its commit; I left them unresolved for you.

| review point | commit |
|---|---|
| overloads were only for backwards compat, drop them | `a1c2b16` |
| "Are these final?" (bumble, serial, bleak); `sequence` default in prose, not the signature; less mutation | `e4e8228` |
| port isn't needed when borrowing | `f6678a1` |
| `winrt`/`bluez` at the same time → a sum type | `eab0363` |
| `SMPClient` generic over the transport | `7bad118` |
| doc slop; kwargs-only bond functions | `7c91b0b` |

The two calls you made first:
- **The link target moves to the primitive.** `connect(target)`/`connected(target)` open a link, `borrow(resource)`/`borrowed(resource)` adopt one, and constructors hold only config.
- **"Less mutation" covers the state this PR introduced.** The Optional `_negotiated_buf_size` is gone; negotiation resolves `Auto` into `BufferSize(n)` in one typed slot. `main`'s pre-existing link fields are unchanged.

Diff against the base: src +1056/−801, tests +812/−484. The src deletions rose because the 7.1.0 params are gone.

